PR #13, opened 2015-10-26
Cp2k

danielhollas:
Merge branch cp2k, which adds direct interfaces to CP2K and TeraChem.


danielhollas:
At this point, TeraChem interface is broken on NEON cluster. Weirdly, it seems to work in Argon cluster. It maybe a TeraChem issue, but it's unlikely, as it worked before.

danielhollas:
I also added initial REMD code. Seems to be working, but not extensively tested.
The whole test suite was updated and should now work everywhere.

PR #35, opened 2020-10-07
Setup code coverage using Codecov.io

### LINKED ISSUE #34: Generate test coverage
Apparently, this is possible by Gfortran, but we will need to write a couple processing scripts and collect the coverage data from all TESTS.

http://www.softeng.rl.ac.uk/st/archive/SoftEng/SESP/html/SoftwareTools/gcov.html

danielhollas:
Closes #34

### DIFF
--- a/.github/workflows/gfortran.yml
+++ b/.github/workflows/gfortran.yml
@@ -89,7 +89,7 @@ jobs:
          gcc_v: [7, 8, 9]
     env:
       FC: mpifort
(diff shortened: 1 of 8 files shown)


danielhollas:
Interesting, the failing basic_build with GCC-9 is failing with a compiler error. :scream: 

Looks like an instance of this:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95847

suchanj (approved):

PR #37, opened 2020-10-12
Clean compiler warnings and fix FFTW and Plumed compilation

danielhollas:
This is mostly removing unused variables. 

Also, previous changes broke the FFTW build.
Plumed build was not broken, but compiling unit tests with Plumed did not work. Now it does.

I also simplified the interface to PRGN a little bit.

### DIFF
--- a/abin.F90
+++ b/abin.F90
@@ -24,12 +24,11 @@ program abin_dyn
    use mod_lz, only: lz_hop
    use mod_kinetic, ONLY: temperature
    use mod_utils, only: abinerror, archive_file, get_formatted_date_and_time
-   use mod_shake, only: nshake
    use mod_transform
    use mod_mdstep
    use mod_minimize, only: minimize
    use mod_analysis, only: analysis, restout
-   use mod_interfaces, only: force_clas, force_quantum
+   use mod_interfaces
    use mod_en_restraint
    use mod_plumed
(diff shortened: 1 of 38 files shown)


danielhollas:
Hmm, the CMD test case is failing, I must have screwed something up. :slightly_frowning_face:

abin.F90
```diff
@@ -61,7 +60,7 @@ program abin_dyn
 
 !  Surface hopping initialization
    if(ipimd.eq.2.or.ipimd.eq.4)then
-      call sh_init(x, y, z, vx, vy, vz, dt)
+      call sh_init(x, y, z, vx, vy, vz)
```
danielhollas, line 63: Whoops, broken in previous commit and was not caught because the `init()` interface was not imported `mod_interfaces`

fortran_interfaces.f90
```diff
@@ -4,10 +4,9 @@ module mod_interfaces
    use mod_const, only: DP
    INTERFACE
 
-   subroutine init(dt, values1)
+   subroutine init(dt)
```
danielhollas, line 7: Whoops, this is an issue from previous commit that was unfortunately not caught because this interface was not imported to `abin.F90`.

suchanj (approved):

PR #39, opened 2020-10-12
Remove on-the-fly histogramming

danielhollas:
Here I remove the somewhat strange ability to print histograms of distances, angles and dihedrals.
Instead, we will print the time series of the values, and histograms can be always done as a post-processing step.


danielhollas:
Oooh, this is pretty great, Codecov! Showing which lines are not covered by tests dirrectly in the PR diff.

obrazek

suchanj (dismissed):

danielhollas:
@suchanj I added unit tests for the functions from this PR. Take a look, they may give you a better idea of how they can look like.

(not saying this is the best way to write them, I am still learning this as well)

PR #41, opened 2020-10-12
Harden TC-MPI connection (handle empty port case)

danielhollas:
This changed was lying in my ARGON repo for a while so I don't remember the context fully,
but apparently I had some issues with TeraChem-MPI interface.

Not sure if this solved the actual issue, but it certainly does not hurt to have it.
We will need to do a thorough testing of this before we compile and release  the new ABIN binary on ARGON.


danielhollas:
Codecov is complaining again that the new code is not tested, but I'll probably merge this and open separate PR for TC-MPI interface tests.

suchanj (approved):

PR #42, opened 2020-10-13
WIP: Add pFUnit library for unit testing

### LINKED ISSUE #32: Create basic unit tests
pFUnit testing framework seems to be maintained and could be nice to use.

https://github.com/Goddard-Fortran-Ecosystem/pFUnit

FRUITPy could be an alternative:
https://github.com/acroucher/FRUITPy

danielhollas:
Closes #32


danielhollas:
The last build with GCC-9 seems to be hitting the --coverage bug again.

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95847
In principle we don't need to collect coverage from all GCC version (just need one) so we should be able to workaround this.

danielhollas:
Forking finally, everything builds, including unit tests with GCC-9, and tests pass.
Tomorrow I'll add some initial documentation about unit testing, but otherwise this is good to go. I can explain the details in person tomorrow.

I had to install pFUnit as part of GitHub CI, and I was able to utilize the build cache in order to significantly speedup subsequent runs once the pFUnit is build and cached. Same approach should probably be utilized for Plumed. See how it is done in `github.yaml`, can explain details in person.

suchanj (approved):

PR #46, opened 2020-11-06
Update Fortran MPI interface: "use mpi" instead of "include mpif.h"

### LINKED ISSUE #33: Use 'use mpi' instead of 'include mpif.h'
Now that we have automated MPI tests, this is a safe change to do.

More info at:
https://github.com/Goddard-Fortran-Ecosystem/pFUnit/issues/34

danielhollas:
Closes #33. This change uncovered a couple of bugs we had in various MPI calls, fortunately nothing critical.

As an unfortunate side-effect, we now need to build our own MPICH and OpenMPI in Github Actions, for each GFortran version,
since `use mpi` does not allow using different GFortran versions for MPI compiler and ABIN. That is unfortunate, but probably a good thing to do anyway. The MPICH and OpenMPI builds take a long time, but we cache them so it's not so bad (GA cache should last at least a week).

As part of this, I also did a bit more cleanup of `Makefile` and `TEST/test.sh`.


WATERMODELS/Makefile
```diff
@@ -1,5 +1,5 @@
 CXX = g++
-CXXFLAGS = -g -O2 -Wall
+CXXFLAGS = -O2 -Wall -std=c++11
```
danielhollas, line 2: This was overlooked in the last PR. :-(

force_tera.F90
```diff
@@ -347,7 +347,7 @@ subroutine connect_terachem( itera )
       call abinerror('force_tera')
    end if
 
-   timer = MPI_WTIME(ierr)
+   timer = MPI_WTIME()
```
danielhollas, line 350: `MPI_WTIME()` does not return any errors. This kind of error can be caught by using `use mpi` instead of `include mpif.h`.

remd.F90
```diff
@@ -143,26 +143,26 @@ subroutine swap_replicas(x, y, z, px, py, pz, fxc, fyc, fzc, rank1, rank2)
...
-      call MPI_recv(y_new, size1*size2, MPI_DOUBLE_PRECISION, dest, tag_y, MPI_COMM_WORLD, status, ierr )
-      call MPI_recv(z_new, size1*size2, MPI_DOUBLE_PRECISION, dest, tag_z, MPI_COMM_WORLD, status, ierr )
-      call MPI_recv(px_new, size1*size2, MPI_DOUBLE_PRECISION, dest, tag_px, MPI_COMM_WORLD, status, ierr )
-      call MPI_recv(py_new, size1*size2, MPI_DOUBLE_PRECISION, dest, tag_py, MPI_COMM_WORLD, status, ierr )
-      call MPI_recv(pz_new, size1*size2, MPI_DOUBLE_PRECISION, dest, tag_pz, MPI_COMM_WORLD, status, ierr )
-      call MPI_recv(fxc_new ,size1*size2, MPI_DOUBLE_PRECISION, dest, tag_fx, MPI_COMM_WORLD, status, ierr )
-      call MPI_recv(fyc_new ,size1*size2, MPI_DOUBLE_PRECISION, dest, tag_fy, MPI_COMM_WORLD, status, ierr )
-      call MPI_recv(fzc_new, size1*size2, MPI_DOUBLE_PRECISION, dest, tag_fz, MPI_COMM_WORLD, status, ierr )
-      call MPI_recv(irank, 1, MPI_INTEGER, dest, tag_fz, MPI_COMM_WORLD, status, ierr )
+      call MPI_Send(x, size1*size2, MPI_DOUBLE_PRECISION, dest, tag_x, MPI_COMM_WORLD, ierr)
```
danielhollas, line 146: Here again, the status parameter should not be there in `MPI_Send()`. Hopefully, this did not introduce any practical problem in the REMD code, since we're not checking the `ierr` variable anyway (but we really should).

TESTS/test.sh
```diff
@@ -3,24 +3,22 @@
 # Parameters passed from Makefile
 ABINEXE="$PWD/$1 -x mini.dat"
 
-# TODO: This is extremely brittle, we should at least
 # verify that the number of parameters is correct!
+if [[ $# -ne 7 ]]; then
```
danielhollas, line 7: Changes in `test.sh` and `Makefile` do not actually pertain to MPI, but they were a nice cleanup.

Makefile
```diff
@@ -94,72 +94,50 @@ ${BIN} : compile_info.o
...
-
-# TODO: Use only .F90
-.F90.o:
+%.o: %.F90
 	$(FC) $(FFLAGS) $(DFLAGS) $(INC) -c $<
 
-.cpp.o:
-	$(CXX) $(CXXLAGS) $(DFLAGS) -c $<
+%.o: %.cpp
+	$(CXX) $(CXXFLAGS) $(DFLAGS) -c $<
```
danielhollas, line 138: Whoops, previous PR had a typo, `CXXLAGS` -> `CXXFLAGS`

Makefile
```diff
@@ -94,72 +94,50 @@ ${BIN} : compile_info.o
...
-.SUFFIXES: .F90 .f90 .f95 .f03 .F03 .cpp
-
-# TODO: Use only .F90
-.F90.o:
+%.o: %.F90
 	$(FC) $(FFLAGS) $(DFLAGS) $(INC) -c $<
 
-.cpp.o:
-	$(CXX) $(CXXLAGS) $(DFLAGS) -c $<
+%.o: %.cpp
```
danielhollas, line 137: This is a more modern approach, as per:
https://www.gnu.org/software/make/manual/html_node/Suffix-Rules.html

Makefile
```diff
@@ -94,72 +94,50 @@ ${BIN} : compile_info.o
...
 	$(MAKE) unittest
 endif
-	/bin/bash TESTS/test.sh ${BIN} $(TEST) ${MPI} ${FFTW} ${PLUMED} ${CP2K}
+	/bin/bash TESTS/test.sh ${BIN} $(TEST) ${MPI} ${FFTW} ${PLUMED} ${CP2K} test
 
 # Clean all test folders.
-testclean :
-	/bin/bash TESTS/test.sh ${BIN} clean ${MPI} ${FFTW} $(PLUMED) ${CP2K}
+testclean:
+	/bin/bash TESTS/test.sh ${BIN} $(TEST) ${MPI} ${FFTW} $(PLUMED) ${CP2K} clean
```
danielhollas, line 125: This is now more consistent.

suchanj (approved):

PR #51, opened 2020-12-19
MAJOR fix to tests!

danielhollas:
Turns out testing have been broken for a while. :-(
Need to find out when, and make sure we didn't break
anything in the meantime. Shit.

@suchanj this is exactly the reason why it's so important to see the test fail. :slightly_frowning_face:

### DIFF
--- a/tests/numdiff.py
+++ b/tests/numdiff.py
@@ -31,8 +31,8 @@ except:
 
 print("Comparing numerical differences in file "+inpfile)
 
-def failed(reference, test):
-   delta = float(reference) - float(test)
+def failed(test, reference):
+   delta = abs(float(reference) - float(test))
    # Let's just not report small differences at all,
    # It just creates a noise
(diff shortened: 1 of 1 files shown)


danielhollas:
So the good news is that the current tests are still passing, even after thi fix.

The bad news is this have been broken for at least two years, perhaps from the very beginning.

The bug would affect the test result only if all the numerical differences had the same sign. That seems statistically unlikely, so there's a cause for cautious optimism.

The lesson is clear: always (always) see the test fail, before it succeeds. Otherwise, it may be that the test is simply unable to fail, and is not effectively testing anything...

https://github.com/PHOTOX/ABIN/commit/893e0b8c9f645971411cafe3c5153c45a288ea93#diff-01d568399d0e5a8e5dc16dd069177417e9c18db8551b5bac9d8d196cd7aa013fL39

suchanj (approved):

suchanj:
Ignore/squash my merge commit, please.

PR #53, opened 2021-01-11
Harden PLUMED restart

danielhollas:
Previously, in order to restart the PLUMED simulations,
the 'RESTART' keyword needed to be added at the beginning
of the PLUMED input file, in addition to setting 'irest=1' in the ABIN input file.

In this diff, we pass the restart info from ABIN to PLUMED directly
so the additional keyword in the PLUMED input is not needed
(but it can still be there).

In addition, I added some basic unit tests for the PLUMED
integration and I improved the current PLUMED test:
1. to test the restart capability
2. print more output files from PLUMED
   to check we're passing correct info from ABIN.
   Specifically, we check masses and energies.

I also added a test for compilation without PLUMED.

Finally, I added the brand new 2.7 PLUMED version to GitHub CI.

Because of these changes, I needed to bump the minimum PLUMED API version.
Nevertheless, the 2.5 version still works. Since 2.5 is the last officially supported version,
I am not worried about compatibility with older versions.

Also, I find out that temperature was passed incorrectly into PLUMED! (it was set to zero). I commented out the call
so PLUMED should now die if temperature is not specified in the plumed input. I'll fix this in a future PR. Having
to specify temperature in two different places is clearly suboptimal.


.github/workflows/gfortran.yml
```diff
@@ -207,7 +207,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-         plumed_v: [2.5.3, 2.6.2]
+         plumed_v: [2.5.3, 2.6.2, 2.7.0]
```
danielhollas, line 210: New version has just been released so I added it here. :-)

.gitignore
```diff
@@ -29,7 +29,6 @@ tests/*/debug.nacm
 tests/*/ERROR
 tests/SH_BUTCHER/geom.?.?.?
 tests/CP2K*/WATER*
-tests/PLUMED/*
```
danielhollas, line 32: This really confused me. Better to have specific rules in .gitignore files.

src/init.F90
```diff
@@ -335,16 +335,7 @@ subroutine init(dt)
    if(ipimd.eq.4) call allocate_ehrenfest(natom, nstate)
 
    if(iplumed.eq.1) then
-
-#ifdef USE_PLUMED
```
danielhollas, line 339: Looks much nice and more readable without ifdefs everywhere. :-)

src/utils.F90
```diff
@@ -136,6 +145,12 @@ function LowerToUpper(string) result (return_string)
       end do
    end function LowerToUpper
 
+   subroutine not_compiled_with(feature, caller)
```
danielhollas, line 148: Just a simple helper function to make our errors uniform and avoid code duplication. Right now only used in PLUMED interface, but I'll use it for MPI and others in future PRs.

unit_tests/Makefile
```diff
@@ -22,24 +23,25 @@ utils_REGISTRY :=
...
 
 clean :
-	/bin/rm -f *.o *.mod *.inc test_*.F90 utils
+	/bin/rm -f bck.* *dat *.o *.out *.mod *.inc test_*.F90 utils plumed
  
 .PHONY: clean test
 
 .SUFFIXES: .F90
-
-.F90.o:
```
danielhollas, line 43: This rule was not actually being used (I think it was being overwritten internally by pFUnit.
So instead I added `$(DFLAGS)` to `$(FFLAGS)` above.

unit_tests/test_plumed.pf
```diff
@@ -0,0 +1,99 @@
...
+      ! Let's move a bit, now the forces should be different.
+      ! We're only changing the x-coordinate
+      ! so fy and fz should be unchanged
+      x(2,1) = 1.5d0
+      it = 2
+      call force_plumed(x, y, z, fx, fy, fz, eclas)
+
+      ! TODO: Make this test work, we need to take a derivative of the Gaussian
+      ! and compute it for deltaX=0.5.
+      !@assertEqual(1.0d0, fx, "fx changed")
```
danielhollas, line 94: Remaining TODO.

src/plumed.F90
```diff
@@ -1,133 +1,200 @@
...
+      integer :: api_version
+      ! Conversion from ABIN to PLUMED units
+      real(DP), parameter :: PLUMED_ENERGY_UNIT = 2625.5697D0    ! Ha -> kJ/mol
+      real(DP), parameter :: PLUMED_LENGTH_UNIT = 0.1D0 / ANG    ! Bohr -> nm
+      real(DP), parameter :: PLUMED_TIME_UNIT = AUTOFS * 0.001D0 ! a.u. -> ps
+      real(DP), parameter :: PLUMED_MASS_UNIT = 1.0D0 / AMU      ! a.u. -> amu
+
+      plumed_kbt = temp * 8.3144598D0 ! in kJ/mol
+
+      write (*, *) 'PLUMED is ON'
```
danielhollas, line 53: Moved here from `init.F90`

src/plumed.F90
```diff
@@ -1,133 +1,200 @@
...
+      ! This needs to be the last call
+      call plumed_f_gcmd(c_string("init"), C_NULL_PTR)
+      ! Make sure PLUMED flushes file buffers with a reasonable frequency
+      ! in case we suddenly die. We set the frequency to every nrest steps
+      ! to ensure we have consistent data for restarting the simulation.
+      write (chint, *) nrest
+      ! NOTE: 'readInputLine' cannot be called before 'init'
+      ! Only available for API_VERSION > 3
+      ! https://www.plumed.org/doc-v2.6/user-doc/html/_f_l_u_s_h.html
+      call plumed_f_gcmd(c_string('readInputLine'), 'FLUSH STRIDE='//c_string(chint))
```
danielhollas, line 98: This is a new thing, PLUMED by default flushes only once per 1000, so if ABIN died unexpectedly (e.g. electrical failure), the data would be lost.

src/plumed.F90
```diff
@@ -1,133 +1,200 @@
...
+   end subroutine plumed_init
+
+   subroutine finalize_plumed()
```
danielhollas, line 101: Just trying to encapsulate any direct calls to PLUMED interface to this `mod_plumed` module.

src/plumed.F90
```diff
@@ -1,133 +1,200 @@
...
+      use mod_utils, only: c_string
+      implicit none
+      real(DP), intent(in) :: x(:, :), y(:, :), z(:, :)
+      real(DP), intent(in) :: eclas
+      real(DP), intent(inout) :: fx(:, :), fy(:, :), fz(:, :)
+      real(DP) :: xx(size(x, 1)), yy(size(y, 1)), zz(size(z, 1))
+      real(DP) :: fxx(size(fx, 1)), fyy(size(fy, 1)), fzz(size(fz, 1))
+      ! Currently not implemented
+      !real(DP) :: pbox(3,3), pcharges(natom)
+      real(DP) :: plumvirial(3, 3)
```
danielhollas, line 117: I wanted to get rid of this, but PLUMED crushes if we don't provide the virial.
On the other hand, I was able to get rid of the charges, which we do not support,
and PLUMED now crashes if a user requests an Action based on charges.

src/plumed.F90
```diff
@@ -1,133 +1,200 @@
...
-   real(DP)    :: fxx(size(fx,1)),fyy(size(fy,1)),fzz(size(fz,1))
-   real(DP)    :: amm(natom)
-   real(DP)    :: pbox(3,3),pcharges(natom),plumvirial(3,3)
-   integer     :: iat, iw
-
-   ! Apply PLUMED only to centroid/single bead in PIMD ("post-quantization restraint" )
-   ! (https://aip.scitation.org/doi/abs/10.1063/1.4986915)   
-   iw = 1
-
-   do iat = 1, natom
```
danielhollas, line 85: We now set mass units in `plumed_init` so this conversion here is not needed anymore. I added a test for this in `tests/PLUMED`

src/plumed.F90
```diff
@@ -1,133 +1,200 @@
...
+      ! We'd enter an infinite loop!
+   end subroutine finalize_plumed
+
+   subroutine force_plumed(x, y, z, fx, fy, fz, eclas)
+      real(DP), intent(in) :: x(:, :), y(:, :), z(:, :)
+      real(DP), intent(inout) :: fx(:, :), fy(:, :), fz(:, :)
+      real(DP), intent(inout) :: eclas
+
+      ! Just to get rid of compiler warnings :-(
+      fx = x; fy = y; fz = z; eclas = 0.0D0
```
danielhollas, line 194: Ideas on how to get rid of this are welcomed. :-)

src/plumed.F90
```diff
@@ -1,133 +1,200 @@
...
+      ! This is just a label.
+      call plumed_f_gcmd(c_string("setMDEngine"), c_string("ABIN"))
+      ! Set PLUMED input file.
+      call plumed_f_gcmd(c_string("setPlumedDat"), c_string(plumedfile))
+
+      call plumed_f_gcmd(c_string("setNatoms"), natom)
+      call plumed_f_gcmd(c_string("setTimestep"), dt0)
+      call plumed_f_gcmd(c_string("setLogFile"), c_string(plumedoutfile))
+
+      if (irest == 1) then
```
danielhollas, line 84: Here's the new thing.

src/plumed.F90
```diff
@@ -1,133 +1,200 @@
...
+      use mod_nhc, only: temp
+      implicit none
+      integer, parameter :: PLUMED_RESTART = 1
+      ! TODO: Maybe make this configurable?
+      character(len=*), parameter :: plumedoutfile = 'plumed.out'
+      character(len=32) :: chint
+      real(DP) :: plumed_kbt
+      integer :: api_version
+      ! Conversion from ABIN to PLUMED units
+      real(DP), parameter :: PLUMED_ENERGY_UNIT = 2625.5697D0    ! Ha -> kJ/mol
```
danielhollas, line 46: I'm not sure how this value was calculated, but I value based on recent NIST data seems to be
`2625.4996394798254`

src/plumed.F90
```diff
@@ -1,133 +1,205 @@
...
+      !real(DP) :: plumed_kbt
+      integer :: api_version
+      character(len=32) :: chint
+
+      ! TODO: Fix this! Currently, we're calling plumed_init()
+      ! before the temperature is ready, so temp=0.0d0 here!
+      ! Also, this conversion only works if temp is in Kelvins,
+      ! but we use temp in atomic units in ABIN!
+      ! Verify kbt using https://www.plumed.org/doc-v2.5/user-doc/html/kt.html
+      !plumed_kbt = temp * 8.3144598D0 ! in kJ/mol
```
danielhollas, line 55: This was a bug, not sure when it was introduced or if this ever worked. This could potentially influenced old well-tempered metadynamics simulations, if the user forgot to specify TEMP in the PLUMED input. In that case, we passed zero temperature here. :-(

danielhollas:
@suchanj this is ready for review. Happy to answer any questions. I ended up doing a general cleanup of `plumed.F90` and I also applied fprettify to it so it's a bit harder to review, sorry about that. That are only a couple of meaningful changes though, I tried to point those out with comments.

In the meantime, I have one remaining TODO in the new unit tests. I also would like to go through all the steps from the debugging PLUMED doc: https://www.plumed.org/doc-v2.7/developer-doc/html/_how_to_plumed_your_m_d.html#debugging-patch
These are the ones I already checked:
- [x] Positions
- [x] Timestep
- [x] Energy
- [x] Masses
The remaining ones are:
- [ ]  Check passed forces
- [ ] Check forces on energy 

I maybe write some more tests in the next PR. I also need to fix the passing of temperature from ABIN to PLUMED, see comments in the code. I'll write a test for well-tempered metadynamics for that.

danielhollas:
I verified that the code coverage looks good. The relative drop in coverage in plumed.F90 is just because of an auxiliary function that is never called when ABIN is compiled without PLUMED.

src/utils.F90
suchanj: Ok, this is clever.

src/modules.F90
```diff
@@ -8,21 +8,36 @@
...
+   ! to the SI redefinition of kilogram in 2019.
+   ! TODO: Define all other constants based on base SI units.
+   real(DP), parameter :: AVOGADRO = 6.02214076D23
+   real(DP), parameter :: PLANCK = 6.626070150D-23
+   ! Hartree to Joul converstion according to
+   ! https://physics.nist.gov/cuu/Constants/Table/allascii.txt
+   real(DP), parameter :: AUTOJ = 4.3597447222071D-18 ! hartree -> J
    real(DP), parameter :: AMU=1822.888484264545d0 ! atomic mass unit 
-   real(DP), parameter :: ANG=1.889726132873d0     ! nagstroms to bohrs
+   real(DP), parameter :: ANG = 1.889726132873D0    ! Agstroms To Bohrs
```
suchanj, line 29: We both droped "n" in the word aNgstrom

src/plumed.F90
suchanj: Yes, I frequently used it because 1000 steps in ab initio world can take ages and meanwhile you cannot work on the data thats already there. I wonder what gets priority if you redefine flush stride in plumed.in. Maybe we could check whether chint (= nrest) > 1000 ... lets wait for the user experience.

src/plumed.F90
suchanj: I think, thanks to the exponential in WTMetaD formula this would induce HILLS with constant or zero height? Anyway good catch.

suchanj (approved):

src/plumed.F90
danielhollas: Good question. I'll test it out.

src/plumed.F90
danielhollas: Yeah, I think you'd have noticed if this has been an issue. But it will be good to have this fixed.

src/modules.F90
danielhollas: Good catch. :-) Thanks for the review!

src/plumed.F90
danielhollas: It seems that PLUMED considers `FLUSH` as any other ACTION so if you provide it more than once, it will presumably flush files more than once (possibly with different frequency). So I think we're good here. By default, `nrest=1` in ABIN so this feels as a good approach. Setting up `nrest` to something much higher doesn't make sense, and even then PLUMED will simply flush at its default pace according to [the doc](https://www.plumed.org/doc-v2.5/user-doc/html/_f_l_u_s_h.html): 

> Notice that all files are flushed anyway every 10000 steps.

src/plumed.F90
```diff
@@ -1,133 +1,231 @@
...
+      ! Only available for API_VERSION > 3
+      ! https://www.plumed.org/doc-v2.6/user-doc/html/_f_l_u_s_h.html
+      call plumed_f_gcmd(c_string('readInputLine'), 'FLUSH STRIDE='//c_string(chint))
+   end subroutine plumed_init
+
+   subroutine finalize_plumed()
+      integer, parameter :: PLUMED_G_INITIALIZED = 1
+      integer :: initialized
+
+      call plumed_f_ginitialized(initialized)
```
danielhollas, line 129: I just added this since I needed it for unit testing, but it's good to have nonetheless. 
Now we can call `finalize_plumed()` safely even before we managed to initalize it (e.g. if ABIN exits early in `init.F90`).
FYI @suchanj

unit_tests/test_plumed.pf
```diff
@@ -0,0 +1,331 @@
...
+   logical, parameter :: SILENT = .true.
+   integer :: iunit
+   save
+
+contains
+
+#ifdef USE_PLUMED
+
+   ! This routine is called automatically before each test.
+   @before
```
danielhollas, line 21: This is very useful. See also `@after` below.

unit_tests/test_utils.pf
```diff
@@ -164,15 +174,17 @@ contains
 
    end subroutine test_shifted_dihedral
 
-   @test(ifdef=SKIPPED_TESTS)
+   @test
+   @disable
```
danielhollas, line 178: An easy way to temporarily disable a unit test.

danielhollas:
Okay, I spent some more quality time with the unit tests for PLUMED and learned a bit more about how pFUnit framework can be used. I'm fairly happy with the result. :tada: 

@suchanj I'm going to merge this, but it would be great if you could take a second look at `unittests/test_plumed.pf` and let me know if it makes sense to you. Also you can try running unit tests locally if you use my `make.vars` on NEON in `/home/hollas/ABIN-DEV`. I tried to make comments and and make the tests easy to follow, not sure if I succeeded. Happy to incorporate any feedback in future PR.

PR #49, opened 2020-12-18
Surface floating-point operation errors in Github CI

danielhollas:
To catch issues like #48. The tests are now failing, as they should, so this should be merged only once the #48 is addressed.

Some context can be find here:
https://docs.oracle.com/cd/E19957-01/805-4940/6j4m1u7pj/index.html

FYI: @suchanj

PR #54, opened 2021-01-12
Fix test.sh

danielhollas:
Unfortunately, when I refactored `test.sh` in #50 to make it more resilient, I broke it. :-(
Specifically, it was always returning 0, even if some tests failed.
In turn, Github CI was not failing when tests where failing. (as seen e.g. [here](https://github.com/PHOTOX/ABIN/runs/1614264744?check_suite_focus=true)
And this is also a reason why the coverage changes in #49 were so wonky...I should have taken a hint.

Apart from fixing the `test.sh`, I'm also fixing the failing tests here. 

1. Use of unitialized variable `equant` in `analysis.F90`  caused any non-Path-Integral test to fail when compiled with
   floating-point checks (uncovered in PR #50).
   Fortunately, this bug did not really affect anything  because we don't currently print equant anywhere.

2. Non-significant numerical difference in the HARMON test,
   which started showing up after a fix to `numdiff.py` in 4620895 in #51.

A lesson here is, never touch `test.sh` again without opening a dedicated PR that can be carefully reviewed.
The error was introduced in PR #50 which unrelated to tests.


danielhollas:
@suchanj  can you take a look?

The coverage change in `init.F90`, which is not modified here at all, are because the failing test were triggering error code paths that we're currently not testing, see: https://app.codecov.io/gh/PHOTOX/ABIN/compare/54/changes#D1L1095

suchanj:
It all makes sense now, good job unraveling this mystery. I agree with dedicated PRs for test.sh. I wish this could be enforced by some security settings, but minding the red flags and co-reviewing should do.

suchanj (approved):

danielhollas:
Yeah, there were multiple red flags that I didn't paid closer attention to, this one is on me.
I actually did try to make the `test.sh` more robust against programming errors by using `set -euo pipefail`, but it clearly cannot catch all errors.  Weird Codecov report is actually a good sign. Especially if it is reporting coverage changes in files that were not touched in a given PR.

PR #55, opened 2021-01-13
Refactor cleaning test directories

danielhollas:
This is a minor refactor of how we clean up test directories in tests/*/.

We clean the test directory each time we run the test. 
We also cleanup all test directories by running `make testclean`.

As part of this, I renamed `mini.dat` to `mini.xyz` in all tests, so that we can simply delete all *.dat files.

I also cleaned up the diff_files and makeref functions,
hopefully they are a bit more readable now. We meed tp
carefully test that I didn't break anything!

To ensure I did not forget any output files, I asked Git about the untracked files after
I ran the `make testclean` command. To find all untracked files, even those ignored by .gitignore, I used `git ls-files -o`

Test plan:
 - clone fresh repo
 - `./configure && make && make test`
 - `make clean && make testclean`
 - `git ls-files -o` should not list any extra files.


danielhollas:
@suchanj can you take a look? I promise this is the last time I am touching `test.sh` for a while. :-) 
When reviewing the changes in `test.sh`, it might useful to hide the whitespace changes in the Github UI.

obrazek

Also, maybe can you look at the whole `test.sh` and let me know if there's anything unclear? Thanks!

.gitignore
```diff
@@ -1,43 +1,6 @@
-*mod
+# See also .gitignore files in doc/, tests/ and src/
```
danielhollas, line 1: I've split the `.gitignore` file into several separate one in respective directories. Feels a bit clearer and there's less danger that we ignore something by mistake.

tests/test.sh
```diff
@@ -24,79 +23,100 @@ ACTION="$7"
 # It's not needed if you use system-wide MPI installation
 # or if mpirun is in your PATH.
 
-if [[ $ACTION = "makeref" ]];then
+if [[ $ACTION = "makeref" && $TESTS = "all" ]];then
```
danielhollas, line 26: This was a logic error introduced in some previous PR.

tests/test.sh
```diff
@@ -24,79 +23,100 @@ ACTION="$7"
...
+  # Compare test results with all existing reference files
+  local reference_files=$(ls *.ref)
+  if [[ -z $reference_files ]];then
+    echo "ERROR: No reference files were found"
+    return 1
+  fi
+
+  for ref_file in $reference_files
+  do
+    test_file=$(basename $ref_file .ref)
```
danielhollas, line 50: No real changes here in this function, just trying to simplify the logic.

tests/test.sh
```diff
@@ -24,79 +23,100 @@ ACTION="$7"
...
-done
-return $status
+function diff_files {
+  local return_status=0
+  local error_code
+  local ref_file
+  local test_file
+  # Compare test results with all existing reference files
+  local reference_files=$(ls *.ref)
+  if [[ -z $reference_files ]];then
```
danielhollas, line 43: Previously, the test passed if there were no reference files. This is safer.

tests/test.sh
```diff
@@ -24,79 +23,100 @@ ACTION="$7"
...
-   file=$(basename $f .ref)
-   if [[ -f $file.ref ]];then
-      mv $file $file.ref
-   fi
-done
+  local ref_file
+  local test_file
+  local reference_files=$(ls *.ref)
+  echo "Making new reference files."
+  if [[ -z $reference_files ]];then
```
danielhollas, line 87: Here again, we fail loudly if no pre-existing reference files were found. And that is good.

tests/test.sh
```diff
@@ -154,13 +174,8 @@ do
...
    cd $dir
 
-   # Always clean test directory
-   # before runnning the test
-   if [[ -f "test.sh" ]];then
-      ./test.sh clean
-   else
-      clean ${files[@]}
-   fi
+   # Always clean the test directory before runnning the test.
```
danielhollas, line 177: I just moved this logic inside the `clean` function.

tests/test.sh
```diff
@@ -170,39 +185,42 @@ do
...
       #./test.sh $ABINEXE 2> /dev/null
       ./test.sh $ABINEXE || true
 
    else
       if [[ -f "velocities.in" ]];then
-         $ABINEXE -v "velocities.in" > output || true
+         $ABINEXE -v "velocities.in" > $ABINOUT || true
       else
-         $ABINEXE > output || true
+         $ABINEXE > $ABINOUT || true
```
danielhollas, line 204: Since we're using the `set -e` BASH settings, we need this syntax `command || true` for commands that are allowed to fail. ABIN needs to be allowed to fail since we purposefully let it fail in some tests.

tests/test.sh
```diff
@@ -170,39 +185,42 @@ do
...
       if [[ -e input.in2 ]];then
-         $ABINEXE -i input.in2 >> output || true
+         $ABINEXE -i input.in2 >> $ABINOUT || true
       fi
    fi
 
    if [[ $ACTION = "makeref" ]];then
 
-      makeref ${files[@]}
+      makeref
```
danielhollas, line 215: `makeref` function actually does not accept any parameters, this was a remnant of previous versions.

tests/test.sh
```diff
@@ -24,79 +23,100 @@ ACTION="$7"
...
+  if [[ -f "test.sh" ]];then
+    ./test.sh clean
+  fi
+  rm -rf $*
+  rm -f *.diff
 }
 
-files=( *-RESTART.wfn* cp2k.out bkl.dat phase.dat wfcoef.dat restart_sh.bin restart_sh.bin.old restart_sh.bin.?? nacm_all.dat minimize.dat geom.mini.xyz temper.dat temper.dat radius.dat vel.dat cv.dat cv_dcv.dat  dist.dat angles.dat dihedrals.dat geom.dat.??? geom_mm.dat.??? DYN/OUT* MM/OUT* state.dat stateall.dat stateall_grad.dat ERROR debug.nacm dotprod.dat pop.dat prob.dat PES.dat energies.dat est_energy.dat movie.xyz movie.xyz movie_mini.xyz restart.xyz.old restart.xyz.? restart.xyz.?? restart.xyz SOC.dat )
+# List of all possible ABIN output files.
+# Used by `make testclean` to cleanup test directories.
```
danielhollas, line 112: Thanks to the `*.dat` this is now a much more manageable list.

suchanj:
Note: after running 'make testclean' the last message is "All tests PASSED." But this is just a minor thing, it would require redefining final section handling the global_error.

suchanj (approved):

danielhollas:
@suchanj thanks. Agreed that final message after `make testclean` is a bit confusing, but I didn't figure out any easy fix without complicating the logic at the end (and I want to make this script as simple as possible).

PR #56, opened 2021-01-17
Encapsulate FFTW, squash more compiler warnings

### LINKED ISSUE #23: PILE tau0
We need to set default tau0 value or warn the user if its not set. Otherwise the program runs with NaN positions.

danielhollas:
A minor refactor of how we call into fast fourier transform routines withing FFTW library. 
Now all such calls are contained withing the `mod_fftw` module in `fftw_interface.F90`.

Notably, I was able to re-enable to PILE test in the Github CI. I needed to disable PIGLET test which is failing.
We need a deeper investigation as to way. Anyway, test coverage got bumped again. :-) 

I also cleaned up more compiler warning across the code.
Those that remained are connected to MPI, I'll tackle them in a separate PR.

In the third commit I also addressed #23.
Previously, we were reusing `tau0` defined in `nosehoover.F90`, which was very confusing.
So I renamed it to `tau0_langevin`. Arguably, we should also rename the other one to `tau0_nosehoover`
to make things consistent, but that would break old input files. Maybe in the future we could support both of them?

Closes #23


src/fftw_interface.F90
```diff
@@ -3,26 +3,23 @@
 module mod_fftw3
    use, intrinsic :: iso_c_binding
    private
-#ifdef USE_FFTW
-   public :: fftw_init, fftw_end, plan_utox, plan_xtou
-   ! TODO: Not sure that these arrays should be public and live here,
-   ! They should probably we allocated where they are used in transform.F90
-   public :: x_in, y_in, z_in, cx, cy, cz
```
danielhollas, line 10: I moved these arrays into `transform.F90`. That's the more logical place for them.
This module should not handle input and output arrays.

src/force_tera.F90
```diff
@@ -471,21 +474,22 @@ subroutine finalize_terachem(error_code)
    end subroutine finalize_terachem
 
    ! TODO: call this after each MPI call
-   subroutine handle_mpi_error(mpi_err, mpi_status, mpi_datatype, expected_count)
+   subroutine handle_mpi_error(mpi_err, status, datatype, expected_count)
+   use mpi
    use mod_utils,    only: abinerror
    integer, intent(in) :: mpi_err
-   integer, intent(in), optional :: mpi_status(MPI_STATUS_SIZE)
```
danielhollas, line 477: MPI_STATUS and MPI_DATATYPE seem to be reserved keywords in the MPI module so I renamed them here.

src/forces.F90
```diff
@@ -185,12 +185,8 @@ subroutine force_wrapper(x, y, z, fx, fy, fz,  e_pot, chpot, walkmax)
    SELECT CASE (chpot)
       case ("mm")
          call force_LJ_Coulomb(x, y, z, fx, fy, fz, eclas)
-#ifndef CP2K
```
danielhollas, line 188: TODO: Need to test this!

src/init.F90
```diff
@@ -125,19 +122,13 @@ subroutine init(dt)
...
-      write(*,*)'ERROR: ABIN was not compiled with CP2K interface.'
-      write(*,*)''
-      call abinerror('init')
-#endif
 #ifdef USE_MPI
    else
       call MPI_INIT ( ierr )
       if (ierr.ne.0)then
          write(*,*)'Bad signal from MPI_INIT:', ierr
-         call abinerror('MPI_INIT')
```
danielhollas, line 140: This is so early in the initialization that most of the things will not be initialized/allocated at this point.
So I thought direct  STOP might be a better approach here.

src/transform.F90
```diff
@@ -1,21 +1,92 @@
...
    use mod_utils, only: abinerror
    implicit none
    private
    public :: UtoX, XtoU, QtoX, XtoQ, FXtoFQ, FQtoFX
-   public :: init_mass
+   public :: init_mass, initialize_pi_transforms
    public :: equant_nm
-   contains
+   public :: finalize_normalmodes
+   real(C_DOUBLE), dimension(:), allocatable :: x_in, y_in, z_in
```
danielhollas, line 15: Moved here from `fftw_interface.F90`

danielhollas:
@suchanj this one's ready for review as well. What remains to be done is to reuse the new `not_compiled_with()` function once I merge the PLUMED PR and rebase this one on top.

It would be nice to add some unit test for the Fourier Transform routines, but I feel this PR has already a lot. I'll try to keep the future PRs smaller than this.

src/force_cp2k.F90
```diff
@@ -1,9 +1,13 @@
 module mod_cp2k
    use iso_c_binding
    use mod_const,    only: DP
+#ifndef CP2K
```
danielhollas, line 4: Feel free to completely ignore this file in the review, I only deleted unused variables and this is untested/unofficial code anyway.

tests/WITHOUT_FFTW/ERROR.ref
```diff
@@ -0,0 +1,4 @@
+ FATAL ERROR encountered in subroutine: fftw_normalmodes_init
```
danielhollas, line 1: This error comes from `input.in`, while the next one comes from `input.in2`.

suchanj (approved):
Ok, looks good to me.

Issue #27, opened 2020-10-01
Stabilize tests


danielhollas:
We should modify tests so that each test is only 1 or 2 MD steps (+restart). That should minimize the differences between compilers/machines so we could be more strict when comparing against reference values.

PR #58, opened 2021-01-20
Tweaks and fixes for GA and Codecov configs

danielhollas:
- Codecov reports with GFortran 8 and 9 were crashing because we were not setting the correct 'gcov' version.

- Some caches were not used correctly.
  The GA config is getting a bit unwieldy. :-(

- GA: Pin Ubuntu version to 18.04:
  Github will soon switch it's 'ubuntu-latest'
  runner to 'ubuntu-20.04' and we're not ready:
  1. Some tests are failing with minor numerical differences on 20.04
     Hopefully this get's solved with #27.

  2. Some jobs in GFortran CI config now rely on the fact
     that the default GCC/GFortran version is 7. We'll need to make that explicit
     if we want to switch to 20.04. We want to keep the default version = 7
     since that's the one we're using on the PHOTOX clusters.

  3. Given our ancient Linux versions on PHOTOX clusters,
     it probably makes sense to stick to 18.04 for now.
     (and maybe in the future switch one job to 20.04)

- GA: Separate unit tests from end-to-end tests
  This is just for a convenience, to make it easier
  to inspect Github Action workflows. This will matter more
  as we'll add more unit tests. :-)
  This also makes the Makefile a bit more consistent;
  treating e2e and unit tests on equal footing.

### DIFF
--- a/.github/workflows/gfortran.yml
+++ b/.github/workflows/gfortran.yml
@@ -5,11 +5,20 @@ on:
     branches: [ master ]
   pull_request:
 
+env:
+  # Options for upload to Codecov
+  # -Z : Fail CI if Codecov upload fails
+  # -v : verbose output
+  # -X : disable unneeded functionality
+  # CURL_OPTS : Extra curl params to avoid timeout issues / for debugging
+  CURL_OPTS:  -S --no-silent
+  CODECOV_OPTIONS: -Z -X coveragepy -X xcode
 
 jobs:
 
   basic_build:
-    runs-on: ubuntu-latest
+    name: Basic build
+    runs-on: ubuntu-18.04
     strategy:
       fail-fast: false
       matrix:
@@ -18,13 +27,17 @@ jobs:
       FC: gfortran
       FFLAGS: -O0 -fopenmp -Wall --coverage -ffpe-trap=invalid,zero,overflow,denormal
       GCC_V: ${{ matrix.gcc_v}}
+      CODECOV_NAME: ${{format('{0} GCC-{1}', github.job, matrix.gcc_v)}}
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 2
     - name: Set GFortran version
       run: |
         sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-${GCC_V} 100 \
-        --slave /usr/bin/gfortran gfortran /usr/bin/gfortran-${GCC_V}
+        --slave /usr/bin/gfortran gfortran /usr/bin/gfortran-${GCC_V} \
+        --slave /usr/bin/gcov gcov /usr/bin/gcov-${GCC_V}
 
     # pFUnit library is used to build and run unit tests
     - name: pFUnit build Cache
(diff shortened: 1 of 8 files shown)


.github/workflows/gfortran.yml
```diff
@@ -24,7 +24,8 @@ jobs:
     - name: Set GFortran version
       run: |
         sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-${GCC_V} 100 \
-        --slave /usr/bin/gfortran gfortran /usr/bin/gfortran-${GCC_V}
+        --slave /usr/bin/gfortran gfortran /usr/bin/gfortran-${GCC_V} \
+        --slave /usr/bin/gcov gcov /usr/bin/gcov-${GCC_V}
```
danielhollas, line 28: Code coverage for non-default GCC (8 and 9) was failing before this change.

.github/workflows/gfortran.yml
```diff
@@ -221,15 +236,17 @@ jobs:
...
 
   plumed_build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
     needs: basic_build
     strategy:
       fail-fast: false
       matrix:
          plumed_v: [2.5.3, 2.6.2, 2.7.0]
+         gcc_v: [7]
```
danielhollas, line 245: Because the GGC_V variable was not specified, we weren't reusing the pFUnit cache.

src/Makefile
```diff
@@ -12,7 +12,7 @@ ${BIN} : libabin.a
 
 # Used for building Unit Tests
 libabin.a: compile_info.o
-	ar cru libabin.a ${C_OBJS} $(F_OBJS) compile_info.o && ranlib libabin.a
+	ar rcs libabin.a ${C_OBJS} $(F_OBJS) compile_info.o
```
danielhollas, line 15: According to `man ranlib`:

> The GNU ranlib program is another form of GNU ar; running ranlib is completely equivalent to executing ar -s

.github/workflows/gfortran.yml
```diff
@@ -18,13 +27,17 @@ jobs:
       FC: gfortran
       FFLAGS: -O0 -fopenmp -Wall --coverage -ffpe-trap=invalid,zero,overflow,denormal
       GCC_V: ${{ matrix.gcc_v}}
+      CODECOV_NAME: ${{format('{0} GCC{1}', github.job, matrix.gcc_v)}}
```
danielhollas, line 30: This is needed to debug missing Codecov reports, otherwise I cannot distinguish in the Codecov UI which jobs uploaded reports successfully.

.github/workflows/gfortran.yml
```diff
@@ -217,32 +260,38 @@ jobs:
...
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 2
     - name: Plumed build cache
       id: plumed-cache
       uses: actions/cache@v2
       with:
         path: ~/plumed/${{ env.PLUMED_V }}/install
-        key: ${{runner.os}}-plumed${{env.PLUMED_V}}-${{hashFiles('dev_scripts/install_plumed.sh')}}
+        key: ${{runner.os}}-plumed${{env.PLUMED_V}}-gcc${{ env.GCC_V }}-${{hashFiles('dev_scripts/install_plumed.sh')}}
```
danielhollas, line 294: To clear the cache in case we decide to change the GCC version in the future.

.github/workflows/gfortran.yml
```diff
@@ -41,19 +54,25 @@ jobs:
...
 
-   # TODO: Separate unit test coverage
-    - name: Codecov upload
-      run:  bash <(curl -s https://codecov.io/bash)
+    - name: Run End-to-End tests
+      run: make e2etest
+
+    # TODO: Separate unit test coverage
+    - name: Upload code coverage to Codecov
+      run:  cd src && bash <(curl -s https://codecov.io/bash) $CODECOV_OPTIONS -U "$CURL_OPTS"
```
danielhollas, line 68: It looks like `gcov` is working better when running directly on the `src/` directory. Before this change some builds were sometimes missing or empty (specifically for GCC-8 and GCC-9 builds).

danielhollas:
@suchanj this is finally ready for review. :tada: 

It took me a long time to stabilize the Codecov behavior, now it seems to do the trick, it seems that the BASH uploader needs to be from within the `src/` directory.

The coverage changes are a bit puzzling, but I verified that all of them are correct, i.e. from the code that we do not currently test. I guess these lines were not picked up with gcov-7, and they showed up now that the coverage reports work when compiling with GCC8 and GCC9 (see comments inline).

.github/workflows/gfortran.yml
```diff
@@ -18,13 +27,17 @@ jobs:
       FC: gfortran
       FFLAGS: -O0 -fopenmp -Wall --coverage -ffpe-trap=invalid,zero,overflow,denormal
       GCC_V: ${{ matrix.gcc_v}}
+      CODECOV_NAME: ${{format('{0} GCC-{1}', github.job, matrix.gcc_v)}}
```
danielhollas, line 30: This is useful when debuggin builds uploaded from codecov. Otherwise, they are not distiguishable. Here's how it looks now, e.g. 

https://codecov.io/gh/PHOTOX/ABIN/commit/2860c44af6aad41e4730f4b9bc75bc126b4f1e36/build

image

suchanj:
Ok, it is getting harder to keep up :D, but I got the gist of it.

suchanj (approved):

PR #62, opened 2021-02-03
Test TeraChem MPI interface

danielhollas:
The primary goal here was to write tests for the TeraChem MPI interface implemented in `src/force_tera.F90`. There's also a separate interface for Surface Hopping defined in `force_terash.F90`, but that will be handled in a follow-up PR. 

I also ended up doing a little refactor and separated some common functions from `force_tera.F90` to `tera_mpi_api.F90`. The advantage is that they can now be more easily reused in  `force_terash.F90` and `force_tera.F90` and `force_terash.F90` are now independent. I also spend quite some time improving the error handling so the interface should now be more robust and easier to debug.

The tricky thing here is of course the TC part. We don't want to run real TC as part of our tests so I essentially ended up extracting the TeraChem side of the API from TeraChem source code and we compile and run it as a fake TC server as part of the tests (see key files below). Moreover, the TC server is compiled  together with the q-TIP4P/Fw force field so it returns real forces and energies (of course only for pure water). I reused the same code that is part of ABIN itself and that we use in other tests (`pot='mmwater'`). Thanks to this trick, I was able to generate the test reference files (`*ref`) by simply running ABIN with qTIP4P force field, and that way I could verify that the fake TC server works correctly and returns correct energies and forces.

There are several tests:

- `tests/TERAPI`: The most common use case, classical MD.
- `tests/TERAPI-PIMD`: Mostly the same as above but with several PI beads. The only difference in terms of the API is that we pass a different scratch directory name for each bead so that each bead has it's separate scratch dir. (the fake TC server does not really populate the scratch directory. Instead, for testing purposes it creates a file with the same name as the scratch directory).
- `tests/TERAPI-PIMD-PARALLEL`: Same as above, but we connect to several TC servers so that we can run different beads in parallel. This is controlled by the `nteraservers` variable in the ABIN input.
- `tests/TERAPI-FAILS` : This is in fact a collection of different tests verifying different failure modes and error handling. Thanks to these tests I was able to achieve almost 100% test coverage (see Codecov report).

There are a lot of new files, but most of them are new test files. In any case, it's probably easier to check out this branch locally instead of browsing it in GitHub UI.

Key files:
 - `src.force_tera.F90` and `src/tera_mpi_api.F90`
 - `tests/tc_mpi_api.cpp`, `tests/tc_mpi_api.h` - TeraChem side of the API
 - `tests/TERAPI*/tc_server.cpp`, the actual code for the server that uses the API and acts as TeraChem (executes in an MD loop).
 - each test is driven by it's own `test.sh` file. Common BASH helper function were extracted to `tests/test_tc_server_utils.sh`

**TODO**

- [ ] Verify that everything works with actual TeraChem, both on Argon and Neon clusters.
- [ ] Verify parallelization over multiple TC servers (with real TeraChem)
- [ ] Figure out how to handle REMD in combination with TeraChem


.github/workflows/gfortran.yml
```diff
@@ -176,7 +176,7 @@ jobs:
          gcc_v: [7, 8, 9]
          mpich_v: ["3.1.3", "3.3.2"]
     env:
-      ABIN_FFLAGS: -O0 -fopenmp --coverage -ffpe-trap=invalid,zero,overflow,denormal
+      ABIN_FFLAGS: -O0 -g -fopenmp --coverage -ffpe-trap=invalid,zero,overflow,denormal
```
danielhollas, line 179: TODO: Revert this change before merging.

.github/workflows/gfortran.yml
```diff
@@ -14,6 +14,9 @@ env:
   CURL_OPTS:  -S --no-silent
   CODECOV_OPTIONS: -Z -X coveragepy -X xcode
 
+  # FFLAGS for building ABIN, applicable for most jobs
+  ABIN_FFLAGS: -O0 -fopenmp -Wall --coverage -ffpe-trap=invalid,zero,overflow,denormal
```
danielhollas, line 18: The only job where we doesn't use this is the `optimized_build`

src/force_tera.F90
```diff
@@ -474,34 +488,18 @@ subroutine finalize_terachem(error_code)
...
+   subroutine handle_mpi_error(mpi_err)
    use mpi
-   use mod_utils,    only: abinerror
+   use mod_utils, only: abinerror
    integer, intent(in) :: mpi_err
-   integer, intent(in), optional :: status(MPI_STATUS_SIZE)
-   integer, intent(in), optional :: datatype, expected_count
-   integer :: received_count, ierr
-
-   if (present(status).and.status(MPI_TAG).eq.MPI_TAG_ERROR)then
```
danielhollas, line 485: For some reason I was getting a segfault on this line when compiling with MPICH and GCC-9, even though we've not been calling this function with the optional `status` input anywhere. Maybe a compiler bug? In any case, since we're currently not using this code I deleted it, and will enable it back in a separate PR.

.github/workflows/gfortran.yml
```diff
@@ -174,9 +179,8 @@ jobs:
       fail-fast: false
       matrix:
          gcc_v: [7, 8, 9]
-         mpich_v: ["3.1.3", "3.3.2"]
+         mpich_v: ["3.3.2", "3.4.1"]
```
danielhollas, line 182: I dropped official support for MPICH versions < 3.3, because the `hydra_nameserver` which we use to connect to TC contained a bug where `MPI_Lookup_name()` would sometimes return invalid port names. See https://github.com/pmodels/mpich/commit/cb6deb3770a3c4a0ca1251190fcef59f2f6fd1f4

dev_scripts/install_mpich.sh
```diff
@@ -17,6 +17,10 @@ TAR_FILE="mpich-${MPICH_VERSION}.tar.gz"
 DOWNLOAD_URL="https://www.mpich.org/static/downloads/${MPICH_VERSION}/${TAR_FILE}"
 INSTALL_DIR="$MPICH_DIR/$MPICH_VERSION/install"
 
+# Github Actions machines have two CPUs, per:
+# https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources
+NCPUS=2
```
danielhollas, line 22: We already do this for OpenMPI build.

dev_scripts/install_mpich.sh
```diff
@@ -30,18 +34,22 @@ curl "$DOWNLOAD_URL" > $MPICH_DIR/$MPICH_VERSION/pkg/${TAR_FILE}
...
+#                      This option is needed for GitHub Actions build, I don't know why
+# --with-namepublisher=pmi
+#         This compiled hydra_nameserver binary, needed for MPI interface with TeraChem
+# For production builds, delete the second line of options.
+#export CFLAGS='-g -O0'
+#--disable-fast --enable-g-option=all \
 ./configure FC=gfortran CC=gcc \
-  --enable-fortran=all --disable-cxx \
-  --with-hydra-pm=pmiserv --with-namepublisher=pmi \
+  --enable-fortran=all \
```
danielhollas, line 46: I deleted `--disable-cxx` since we need to `mpic++` to build the mock TeraChem server for testing.

I also added the MPICH interface explicitly `--with-device=ch3` since v3.4.1 uses a new `ch4` default which I'm not sure whether it works. The device is basically an underlying communication protocol that MPICH uses, and can influence performance. But we don't really care about performance that much here, _ab initio_ computation is the bottleneck, MPI communication overhead should be insignificant.

src/force_abin.F90
```diff
@@ -118,12 +118,11 @@ subroutine force_abin(x, y, z, fx, fy, fz, eclas, chpot, walkmax)
...
 
+      ! TODO: Verify that the below is true even for GCC >= 7
       ! Exit status 0 turns to 0
       ! For some reason, exit status 1 turns to 256
-      ! However, this one we get by default from bash, don't know why...
-      ! see this thread for explanation:
-      ! http://coding.derkeiler.com/Archive/Fortran/comp.lang.fortran/MAXUNITS07-01/msg00085.html
-      ! If the bash script wants to notify ABIN, it can use e.g. exit 2
+      ! However, this one we get by default from BASH, I don't know why.
+      ! If the BASH script wants to notify ABIN, it can use e.g. exit 2
```
danielhollas, line 125: This link doesn't work anymore.

src/force_tera.F90
```diff
@@ -278,6 +285,10 @@ subroutine receive_tera(fx, fy, fz, eclas, iw, walkmax, newcomm)
       write(6,'(a,4es15.6)') 'Received QM  dipole moment from server:', dipmom(:,1)
       call flush(6)
    end if
+   if (modulo(it, nwrite) == 0) then
+      call print_dipoles(dipmom(:,1), iw, 1)
```
danielhollas, line 289: We already print this in the Surface Hopping TC interface so I added it here (and the charges above) so that I can test it.

tests/TERAPI-PI/forces.xyz.ref
```diff
@@ -0,0 +1,14 @@
+           3
```
danielhollas, line 1: A bit strange that we don't separate beads here, which is inconsistent with `movie.xyz`.

tests/tc_mpi_api.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+  // NOTE: In the real TC interface, gradients are sent
+  // conditionally only if they are requested.
+  // But we always request them in ABIN (see tc.receive).
+  sendQMGradients();
+}
+
+// Gradients stored internally for now,
+// returns energy in atomic units.
+double TCServerMock::getWaterGradients() {
```
danielhollas, line 260: Basically copied from `src/water_interface.cpp`

src/abin.F90
```diff
@@ -68,7 +68,7 @@ program abin_dyn
 
 !$ nthreads = omp_get_max_threads()
    if (my_rank.eq.0)then
-!$    write(*,*)'Number of OpenMP threads used = ', nthreads
+!$    write (*, '(A,I0)') 'Number of OpenMP threads: ', nthreads
```
danielhollas, line 71: I didn't know I can use variable integer length with format `I0`. That's absolutely awesome, 
I will use it from now on, it will certainly simplify some code...

.github/workflows/gfortran.yml
danielhollas: Btw: That fix is a text-book example of a bug in handling strings in C code. `strlen()` returns length of the string without the terminating null character :-)

danielhollas:
@suchanj feel free to spend as much or as little time on this. :-) I tried to write up the general gist in the PR description, feel free to ask if anything is not clear.

suchanj:
What a chonky PR! Description nicely sums it up, a great step forward.

suchanj (approved):

danielhollas:
There's a remaining TODO related to REMD, but I'll fix that in a separate PR.

PR #64, opened 2021-02-06
Prettify everything!

danielhollas:
The day has finally come to make our code looks nice and consistent. :tada: 

I ran `fprettify` for all files in `src/`, except `fftw3.F90` and `random.F90`
(for the latter the prettification fails hard for some reason).

I also manually modified whitespace around comments
(though not consistently across all files).
I also tried to enforce the maximum-100-chars-per-line rule,
but in some files there were too many so I gave up
and in the end I rerun `fprettify --line-length=140`
to get rid of remaining warnings.
I'll give it another try after some more code refactoring.

Until this is merged on master, the code style defined in `.fprettify.rc` is up to discussion. :-)
In some cases I think there's a bit too much whitespace (e.g. around write and read statements)
but overall I'm happy with the result.

I'm also happy to switch to the C-style comparison operators,
which will be much more friendly for non-Fortran programmers.

Note that there might be some coverage changes here,
since I converted a couple of one-line ifs to multiline.
Let's see.

:exclamation:  NOTE to myself :exclamation: this PR must be merged, not rebased,
since we need to keep the SHA of the prettify commit, see the second commit for explanation.


.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+060f499404c1f7ced91f22a5e86fa2589467d00a # fprettify all source code
```
danielhollas, line 1: This allows for ignoring this commit in the Git history (when using `git blame` command`. See `CONTRIBUTING.md` for explanation.

src/abin.F90
```diff
@@ -229,91 +228,85 @@ program abin_dyn
...
-!use mod_system, only: clean_prop_tempfiles
-
-   ! TODO: Implement "clean" bash function in abin interfaces
-   ! that should be called here (and only if irest=0)
-   call system('rm -f ERROR engrad*.dat.* nacm.dat hessian.dat.* geom.dat.*')
-!   call clean_prop_tempfiles()
+contains
 
-end subroutine clean_temp_files
+   subroutine clean_temp_files()
```
danielhollas, line 306: This is an actual change, I moved this subroutine to inside the `program` statement. This allows it to have explicit interface so we can be sure that this function is called correctly.

codecov.yml
```diff
@@ -27,7 +27,7 @@ coverage:
     patch:
       default:
         target: auto
-        threshold: 1%
+        threshold: 2.0%
```
danielhollas, line 30: Not going to fight with Codecov just because I split some lines that were not covered by tests.

suchanj (approved):

PR #67, opened 2021-05-05
Fix GitHub CI

danielhollas:
Apparently, GCC-8 has been removed from the
Ubuntu-18.04 base image in favour of GCC-10
so we need to adjust.

We still use GCC-7 since this is what is available
on PHOTOX clusters. That's also why we cannot switch
to Ubuntu-20.04 image as it only provides GCC-9/10

I also needed to tweak the FFLAGS for MPI compilation
with GCC-10. The new GFortran does not like when different
types are passed into the same functions (which is typical
for MPI_Send/MPI_Recv).

Lastly, I tried to fix the misbehaving Zenodo badge in README.

NOTE: The Codecov changes here correspond to the previous PR (TC-MPI), since the merge on master failed the tests.

### DIFF
--- a/.github/workflows/gfortran.yml
+++ b/.github/workflows/gfortran.yml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-         gcc_v: [7, 8, 9]
+         gcc_v: [7, 9, 10]
     env:
       FC: gfortran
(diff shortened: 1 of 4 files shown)


danielhollas:
@suchanj can you take a look? Thanks!

suchanj (approved):

PR #70, opened 2021-09-27
Fix compilation with ifort

danielhollas:
These are all ugly hacks for now, but I don't have time to fix these things properly for now.
Long term we need to get rid of all GNU/GCC-specific stuff altogether.

### DIFF
--- a/dev_scripts/setup_dev_environment.sh
+++ b/dev_scripts/setup_dev_environment.sh
@@ -11,14 +11,16 @@ set -euxo pipefail
 CWD=`dirname $0`
 if [[ ! -f $CWD/../.git/hooks/pre-push ]];then
     cp $CWD/pre-push.hook $CWD/../.git/hooks/pre-push
-    chmod +x ../.git/hooks/pre-push
+    chmod +x $CWD/../.git/hooks/pre-push
 fi
 
 # Heuristic: if .vimrc exists, we presume we're dealing with a VIM user :-)
-if [[ -f /home/$USER/.vimrc ]];then
+if [[ -f ~/.vimrc ]];then
   $CWD/setup_vim.sh
 fi
 
 # Install fprettify to autoformat our code.
 # (needs Python 3)
-pip3 install --upgrade fprettify
+# NOTE: Due to a vast variety of different Python environments,
+# you're required to do this step yourself.
+#pip3 install --upgrade fprettify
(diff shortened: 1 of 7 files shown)


dev_scripts/setup_dev_environment.sh
```diff
@@ -11,14 +11,16 @@ set -euxo pipefail
 CWD=`dirname $0`
 if [[ ! -f $CWD/../.git/hooks/pre-push ]];then
     cp $CWD/pre-push.hook $CWD/../.git/hooks/pre-push
-    chmod +x ../.git/hooks/pre-push
+    chmod +x $CWD/../.git/hooks/pre-push
 fi
 
 # Heuristic: if .vimrc exists, we presume we're dealing with a VIM user :-)
-if [[ -f /home/$USER/.vimrc ]];then
+if [[ -f ~/.vimrc ]];then
```
danielhollas, line 18: Durham clusters don't have `/home/$USER/` directory. :-)

src/init.F90
```diff
@@ -1107,6 +1123,8 @@ subroutine print_logo()
       print '(a)', ' '
    end subroutine print_logo
 
+end subroutine init
```
danielhollas, line 1126: `print_runtime_info` has a system call to `ldd`. The fortran compiler is not happy with the fact that we also have a namelist called `system`. Putting `print_runtime_info()` outside of `init()` fixes that.

water_potentials/Makefile
```diff
@@ -1,4 +1,3 @@
-CXX = g++
```
danielhollas, line 1: This should not be hardcoded, but defined in `make.vars`

danielhollas:
@suchanj can you take a look? btw: Big congrats on the poster prize! :tada: 

I think I'll be submitting more PRs in the next few weeks. LMK if you're down for reviewing them, or whether I should merge them without review. Thanks!

suchanj:
@danielhollas Thanks! Regarding the new PRs - I'd like to stay in the loop but don't want to hinder the merging process by taking time. So I would propose that you merge whatever you need, but do not immediately delete the branches from repository so they would still be visible :+1:

suchanj (approved):

danielhollas:
@suchanj awesome, thanks! Sound good. I think some of them will be just some minor refactoring and cleanups. For anything bigger I'll always wait for you to take a look.

PR #72, opened 2021-10-03
Refactor error handling

danielhollas:
Here we tentatively refactor the error handling in ABIN, with the ultimate goal of replacing `abinerror`. 

The main purpose of this refactor is to allow the unit tests to test for error conditions. This is currently not possible since calling `abinerror()` unconditionally stops the program, thus disrupting the test suite. To solve that we add a bit of misdirection in `src/error.F90`. Essentially, we allow pFUnit to swap out our error handler subroutine with its own, which then allows checking assertions in unit tests. This swapping occurs via the `set_error_method` in `unit_tests/throw_with_pfunit.F90`.

The new default error handler is called `fatal_error()` and should be used instead of `abinerror`. I converted only a couple of instance for now, and if this all work, I will convert the rest in a separate PR. 

As an added bonus, the new fatal_error takes different arguments from `abinerror`. It takes the file name, line number, and error message. The first two arguments can be conveniently provided via preprocessor-defined constants `__FILENAME__` and `__LINE__`. This way, we can provide the exact location in the code where the error occurred (and the cost of slightly more verbose code). That's pretty cool! :tada:  I am open for discussion about how the errors should be phrased exactly so that it's clear. The format for now resembles how e.g. gfortran prints errors.

### DIFF
--- a/src/init.F90
+++ b/src/init.F90
@@ -1199,13 +1199,6 @@ subroutine finish(error_code)
    end if
 #endif
 
-   if (my_rank == 0) then
-      write (*, *) ''
-      if (error_code == 0) then
-         write (*, '(A)') 'Job finished!'
-      end if
-   end if
-
    call deallocate_arrays()
 
    ! TODO: Move this to a subroutine in mod_files
--- a/unit_tests/Makefile
+++ b/unit_tests/Makefile
@@ -1,4 +1,5 @@
 # Makefile for ABIN Unit Tests
+#
 # Should be invoked from top level Makefile,
 # which defines certain variables used here,
 # namely path to PFUNIT library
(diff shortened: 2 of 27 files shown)


src/init.F90
```diff
@@ -1199,13 +1199,6 @@ subroutine finish(error_code)
    end if
 #endif
 
-   if (my_rank == 0) then
```
danielhollas, line 1202: I just moved this bit to `abin.F90` where it naturally belongs.

unit_tests/Makefile
```diff
@@ -19,18 +20,26 @@ endif
 # NOTE: $(LIBS) are defined in the root Makefile
 
 utils_TESTS := test_utils.pf
-utils_REGISTRY :=
-utils_OTHER_SOURCES :=
 utils_OTHER_LIBRARIES := -lgcov -fprofile-arcs -L"../src/" -labin -L"../water_potentials/" -lwater ${LIBS}
-utils_OTHER_INCS :=
+utils_OTHER_SRCS := throw_with_pfunit.F90
```
danielhollas, line 24: This is just some pFUnit dark magic boiler plate to make this all work.

suchanj:
Ok, definitely dark magic code. However, brilliant idea.

suchanj (approved):

PR #77, opened 2021-11-22
Ensure implicit none via compiler option

### LINKED ISSUE #63: Compilation with optimalization points to use of uninitialized variables
We should check whether the Warnings are valid (at least those in `transform.F90` seems to be.

### LINKED ISSUE #76: Use `--implicit-none` compiler option
This option should be a default in the Makefile or in the configure - > make.vars.
This would mean we will automatically catch any code that uses implicit typing in Github Actions in the future.

danielhollas:
Adding `-fimplicit-none` to FFLAGS ensures that we always have `implicit none` in our modules and functions.
I fixed one missing `implicit none` (caught an unitialized variable!) and converted one last implicit typing in `random.F90`.

l also moved `mod_const` from `modules.F90` to a separate file `constants.F90` to break circular dependency due to the 'DP' constant. This allowed me reorder compilation of a some files and make 'fatal_error()' accessible to all modules. In general, our goal should be to have one module per file, so while I was in there, I also did the same for `mod_files`.

Other small changes were done to squash compiler warnings. I also did a bit of cleanup in `mass_init` and wrote some unit tests to verify this functionality (especially user defined masses code was messy and not fully tested).

Closes #76, Closes #63

### DIFF
--- a/configure
+++ b/configure
@@ -43,12 +43,14 @@ OPTIONS
 "
 }
 
+warning_flags="-Wall -Wno-integer-division -Wno-maybe-uninitialized"
+
 # Overwrite default options from command line
 PARAMS="$@"
 while [[ $# -gt 0 ]]; do
     case "$1" in
       --debug)
-        FFLAGS="-g -O0 -fopenmp -Wall"
+        FFLAGS="-g -O0 -fopenmp -fimplicit-none $warning_flags"
         shift
         ;;
       --mpi)
@@ -109,7 +111,7 @@ done
 
 # Set to default if FFLAGS is not defined yet, see
 # https://stackoverflow.com/questions/11362250/in-bash-how-do-i-test-if-a-variable-is-defined-in-u-mode
-: ${FFLAGS:='-O2 -fopenmp -Wall'}
+: ${FFLAGS:="-O2 -fopenmp -fimplicit-none $warning_flags"}
 
 if [[ -z ${FC-} ]];then
    if [[ $MPI = "TRUE" ]];then
--- a/src/constants.F90
+++ b/src/constants.F90
@@ -1,0 +1,38 @@
+! Exact values for ANG,AUTOKCAL and AUTODEBYE,me,AUTOM,AUTOFS,AMU:
+! Mohr, Taylor, Newell, Rev. Mod. Phys. 80 (2008) 633-730
+! and using the thermochemical calorie (1 cal = 4.184 J):'
+
+! TODO: Update physical constant according to the latest data
+! https://physics.nist.gov/cuu/Constants/bibliography.html
+
+! ABIN uses atomic units internally
+module mod_const
+   implicit none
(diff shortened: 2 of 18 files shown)


src/files.F90
```diff
@@ -0,0 +1,207 @@
+! Module for permanent file handling
```
danielhollas, line 1: No changes here, just moved from module.F90

src/random.F90
```diff
@@ -542,8 +542,11 @@ end subroutine xuwarm
 
 
       FUNCTION R1MACH()
-      IMPLICIT real(DP) (A-H,O-Z)
-      PARAMETER (ONE=1.D0,TWO=2.D0,HALF=0.5D0)
+      real(DP), parameter :: ONE=1.D0, TWO=2.D0, HALF=0.5D0
```
danielhollas, line 545: Last `implicit real` :-)

src/constants.F90
```diff
@@ -0,0 +1,38 @@
+! Exact values for ANG,AUTOKCAL and AUTODEBYE,me,AUTOM,AUTOFS,AMU:
```
danielhollas, line 1: No changes, just moved from `modules.F90`

src/modules.F90
```diff
@@ -120,29 +83,7 @@ module mod_system
    ! cannot use this
 !   use mod_utils, only: abinerror
    implicit none
-   ! TODO: move more stuff into Properties derived type
-   ! https://kiwi.atmos.colostate.edu/fortran/docs/fortran2012.key-8.pdf
-   ! TODO: We should probably move this outside of the modules.f90
-
-   ! Extra electronic properties (not mandatory)
```
danielhollas, line 127: These were just stubs, not used anywhere, so I removed them for now.

src/modules.F90
```diff
@@ -372,24 +289,58 @@ subroutine mass_init(masses, massnames)
...
-            print*,'I hope you specified the mass in namelist "system"'
+            if (my_rank == 0) then
+               print*,'Atom name ', names(i), ' was not found in the library.'
+               print*,'Using user-defined mass from input file'
+            end if
          end select
       end do
 
-      call init_usermass()
+      ! Check for duplicate user defined atom names
```
danielhollas, line 299: I inlined `init_usermass` for simplicity

configure
```diff
@@ -43,12 +43,14 @@ OPTIONS
 "
 }
 
+warning_flags="-Wall -Wno-integer-division -Wno-maybe-uninitialized"
```
danielhollas, line 46: To remove some spurious compiler warnings. We should now be warnings-free with this setting I think.

danielhollas:
@suchanj thanks for looking at the other PR. :bowing_man:  This one is now ready as well, but absolutely no pressure, do on your own time. Thanks! This is a mixture of several small things, see first comment.

suchanj (approved):
Looking good!

PR #78, opened 2021-11-22
Pin pFUnit version

danielhollas:
pFunit build started to fail with internal compiler error for GCC-7, see details in 

https://github.com/Goddard-Fortran-Ecosystem/pFUnit/issues/274#issuecomment-975740321


danielhollas:
@suchanj FYI merging this without review just to fix the build on Github.

PR #80, opened 2022-02-04
FIX TC Surface Hopping interface, broken since #62

danielhollas:
While the Amber MPI interface that we use for classical MD
expects coordinates in angstromgs, the FMS interface expects bohr,
which I didn't notice when I was refactoring this code.
I really need to write those tests for the TC-SH interface.


suchanj (approved):
Tested, coordinates are passed to TC just right.

PR #81, opened 2022-02-10
Add a basic test for TC-MPI surface hopping interface

danielhollas:
Here's a basic test for the TeraChem MPI interface for Surface Hopping. I verified that this new test would have caught the bug that was fixed in #80. 

As with the previous TC-MPI tests, the new C++ code was copied from TeraChem and trimmed and cleaned up, and I added some basic assertions about what is passed by ABIN. Specificically, in the SH interface, MO coefficients, CI vectors and some other stuff (called 'blob') is passed back and forth so I am checking that the the numbers match. I am also testing the restart capability here. In this interface, there's an extra file `wfn.bin` which stored the MOs and CI vectors per above.

While the whole interface is tested, I am not actually running non-adiabatic dynamics since that would require implementing some kind of a model for NAC vectors. For this test, we're simply running ground state MD, forces and energies are taken from the in-built water force field, as in previous tests. 

As a future todo, we should indeed test a proper surface hopping run. Would be a good opportunity if somebody wanted to take a crack at implementing some of the analytical models for NA dynamics, e.g. one of the 1D Tully models, or vibronic coupling model. Might be a good learning experience for some student. :-)

I have also added some unit tests, mostly to test various error conditions. I checked on Codecov that most of the code in `force_terash.F90` is now tested. There are some lines specific for Landau-Zener, so an additional test will be needed for that in the future.

@suchanj can you take a look? Probably no need to go line-by-line in the C++ code, but hopefully you get the general idea here. Let me know if I can clarify anything.

### DIFF
--- a/src/force_terash.F90
+++ b/src/force_terash.F90
@@ -362,7 +362,7 @@ contains
       call set_nbf(bufints(2))
       call set_blob_size(bufints(3))
 
-      print*, 'size of CI vector, number of AOs, blob size:', bufints(1), bufints(2), bufints(3)
+      print*,'size of CI vector, number of AOs, blob size:', bufints(1), bufints(2), bufints(3)
       call allocate_tc_arrays(nstate, natom)
    end subroutine init_terash
 ! USE_MPI
@@ -435,7 +435,7 @@ contains
       use mod_general, only: it, sim_time, narchive
       use mod_sh_integ, only: nstate
       use mod_utils, only: archive_file
-      character(len=*), parameter :: fname='wfn.bin'
+      character(len=*), parameter :: fname = 'wfn.bin'
       logical :: file_exists
       integer :: uwfn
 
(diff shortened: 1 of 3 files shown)


src/force_terash.F90
```diff
@@ -500,12 +537,6 @@ subroutine move_new2old_terash
       blob_old = blob
    end subroutine move_new2old_terash
 
-   subroutine move_old2new_terash
```
danielhollas, line 503: This was not used anywhere

src/force_terash.F90
```diff
@@ -400,98 +432,103 @@ subroutine finalize_terash()
...
+      use mod_utils, only: archive_file
       use mod_sh_integ, only: nstate
-      character(len=200) :: chout
+      character(len=*), parameter :: fname='wfn.bin'
       logical :: file_exists
       integer :: temp, temp2, time_step
+      integer :: uwfn
       real(DP) :: stime
 
-      if (iremd == 1) then
```
danielhollas, line 447: @suchanj do you know whether there was ever a point where we wanted REMD together with this Surface Hopping interface? I can't imagine a use case for this so I removed this REMD-specific code. I don't think it would have worked anyway.

suchanj: I don't, the idea seems pretty wild, lets remove it.

suchanj (approved):

PR #87, opened 2022-03-12
Remove stubs and comments for Ehrenfest method

danielhollas:
Ehrenfest method was never fully implemented so removing for now to simplify and cleanup the code.

If somebody wants to implement this in the future, this PR will be a good reference point.


src/surfacehop.F90
```diff
@@ -1108,25 +1051,4 @@ integer function check_CIVector(CIvecs, CIvecs_old, ci_len, nstates)
       return
    end function check_CIVector
 
-   ! Choose initial state according to oscillator strength
-   subroutine choose_initial_state(itrj)
```
danielhollas, line 1112: @suchanj  did you ever used this functionality to select initial state based on oscillator strength in surface hopping?

 I don't think this is a good place to have it, might be better to do this outside of ABIN in some driver script.

suchanj: I never did, agree with the driver script.

PR #88, opened 2022-03-12
Reenable GLE and PIGLE tests

danielhollas:
PIGLE and GLE tests were disabled because they were failing in GHA with rather larger numerical differences even though they were fine locally. I have found that the GLE thermostat numerics is extremely numerically sensitive and results generally differ between compiler versions. Not quite sure why that is, and perhaps we could help it a bit if we tried to more carefully implement matrix exponentiation and cholesky decomposition. In this PR, I am taking a more pragmatic approach.

1. I refactored `tests/numdiff.py` to allow specifying per-test numerical tolerance. (ideally, we'd have per-file tolerance, but let's leave that for the future). The tolerance is defined in an optional file `NUM_THRE` in a given test directory.
2. I changed the GLE (renamed to QT) and PIGLE tests to only do two steps, and decreased time step. For PIGLE it was also important to decrease `nabin`. Unfortunately, this alone was not enough. This setup is currently in two new tests QT2 and PIGLE2, which require very high absolute tolerance.
3. The main issue here is that the propagator matrices T and S computed in `compute_propagator()` from GLE-A and GLE-C have numerical differences between compilers, and the errors quickly propagate during dynamics. To get around this, I precomputed these matrices, saved them to binary files GLE-T.bin and GLE-S.bin. In the actual tests, the matrices are then read from these files instead of being computed, which hides the compiler differences. Not an ideal solution, but it works. This is implemented in QT and PIGLE tests (turned on by `gle_test=.true.` switch in `input.in`.
4. PIGLE test was also throwing floating point error because part of the `ak` array in force_quantum was not initialized. I fixed that.

PIGLET test currently fails for me even locally, and I am not sure when it got broke or on which machine it worked. I need to test PIGLET carefully before I enable this test.

TODO:
- [x] finish unit tests
- [x] improve the GLE header, add more citations and DOI links
- [x] get rid of ak array in force_quantum
- [x] handle inose == 4 in init and finish


danielhollas:
Okay, this is extremenly frustrating, I am getting different numerical noise depending on the compiler version. Tried decreasing time step, removing restart, nothing helps. It seems that the instability comes from the propagator matrices, that are computed from A and C matrices read from GLE-A and GLE-C. To get around this this, we could instead try to reading the S and T matrices directly from precomputed files, instead of computing them from A and C. Maybe that will work.

Such sensitivity is worrying. It's possible that our routines for Cholesky decompositiong and matrix exponential are not great.

suchanj:
Thats really interesting different compilers give different errors. I was also surprised Fortran does not have any standard matrix exponential and Cholesky procedure. Nevertheless, solved elegantly for the time being.

suchanj (approved):

danielhollas:
@suchanj thanks! Well, Fortran doesn't even have standard library, unlike Python or C++. The standard way for doing more complex stuff with matrices would be to use BLAS / LAPACK. Here's a Cholesky in Lapack https://www.netlib.org/lapack/explore-html/d1/d7a/group__double_p_ocomputational_ga2f55f604a6003d03b5cd4a0adcfb74d6.html. However, I don't want ABIN to depend on Lapack, because it's always a pain to compile.

I am also not sure it would help, I suspect that the numerical instability is inherent to the numerical procedure itself. I know that in i-Py they switched from Cholesky decomposition to something more stable, so perhaps that's the way to go in the future. (there is a TODO in the source code for this already).

In the meantime I will do a bit more testing to ensure that different compiler versions produce the same thermodynami averages, which is what actually matters in the end.

suchanj:
@danielhollas Alright, I agree we don't want to introduce BLAS / LAPACK dependance, thanks for the extra info!

PR #91, opened 2022-03-18
Refactor GLE module, re-enable PIGLET test

danielhollas:
While trying to re-enable the PIGLET test, my testing showed that PIGLET method did not work. 
After quite some time I found out that PIGLET would only work if `nabin=1` was set explicitly
in the input file, even though PIGLET does not use RESPA integrator,
because nabin is used to determine the timestep that is passed to the gle_init() subroutine. :facepalm: 
Overall, the stuff with all the integrators and thermostat definitely could use some refactor in the future.

Other minor things that I've done while chasing this bug:
- refactored GLE module, especially the `gle_init()` function is now much cleaner. Add new unit tests, this module is now covered almost 100% by tests.
- made internal GLE vars private, made routines for reading / write GLE part of the restart.
- removed the `readQT` restart option, which I don't think make much sense.
- GLE restarts now seems to work with Intel compiler (tested with intel/2018)
- added a LANGEVIN test, classical MD with langevin thermostat (same code path as the PILE thermostat)
- In [this commit](https://github.com/PHOTOX/ABIN/pull/91/commits/d5b8ed073c90bb2db53aeffdc65930da94ddaa3f) I've fixed the long-standing annoyance that we were allocating arrays as `x(natom, nwalk+1)`, where the extra slot was used in some path integral routines
- And finally, colored output from `test.sh`, yay! :tada: Look at this beauty :heart_eyes: 

image

In all seriousness, when a test fails, it is not much easier to visually tell which one failed.


danielhollas:
Did a quick test, and PIGLET does not match PI+GLE results. I am not sure if it ever worked or not. I need to test PILE next. Working PILE would mean that at least normal modes transformation and integration works.

suchanj (approved):

PR #97, opened 2022-03-29
Create MPI wrapper module

danielhollas:
The idea here is to centralize certain calls to MPI library in one place, with proper error handling.
This should reduce the amount of `#ifdef USE_MPI` throughout the codebase.

We also encapsule MPI initialization and finalization, and provide helper functions for error handling.

### DIFF
--- a/.github/workflows/gfortran.yml
+++ b/.github/workflows/gfortran.yml
@@ -231,6 +231,7 @@ jobs:
   openmpi_build:
     name: OpenMPI build
     runs-on: ubuntu-18.04
+    timeout-minutes: 30
     needs: basic_build
     strategy:
       fail-fast: false
--- a/src/init.F90
+++ b/src/init.F90
@@ -673,6 +673,8 @@ contains
 
    subroutine check_inputsanity()
       use mod_chars, only: chknow
+      use mod_utils, only: real_positive, real_nonnegative, &
+                         & int_positive, int_nonnegative, int_switch
       integer :: error
 !$    integer :: nthreads, omp_get_max_threads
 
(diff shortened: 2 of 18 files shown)


.github/workflows/gfortran.yml
```diff
@@ -231,6 +231,7 @@ jobs:
   openmpi_build:
     name: OpenMPI build
     runs-on: ubuntu-18.04
+    timeout-minutes: 30
```
danielhollas, line 234: I noticed lately that sometimes the OpenMPI job in Github Actions hangs when running the tests. I haven't been able to figure out why, since it seems to occur randomly. As a stopgap measure I add here a timeout so that we don't needlessly consume resources.

src/init.F90
```diff
@@ -147,72 +142,39 @@ subroutine init(dt)
...
 
    ! We need to connect to TeraChem as soon as possible,
    ! because we want to shut down TeraChem nicely in case something goes wrong.
-   if (pot == '_tera_' .or. restrain_pot == '_tera_') then
+   if (pot == '_tera_' .or. restrain_pot == '_tera_' .or. pot_ref == '_tera_') then
       call initialize_terachem_interface(trim(tc_server_name))
    end if
 
-   if (en_restraint >= 1) then
-      call en_rest_init()
```
danielhollas, line 204: I noticed that this init function is called before the `natom` is read from the XYZ file. So  I think this would fail unless natom was provided in the input file. I mode the call later in the init. It would be nice to have a test for this functionality. :blush:

suchanj: Whops, noted!

suchanj (approved):

PR #100, opened 2022-03-30
Redirect stdout to /dev/null for my_rank>0

danielhollas:
I was getting increasingly annoyed in how we handle output to stdout, especially when I was writing unit tests and needed to suppress it. For REMD, we only want the first replica to write to stdout, so we had a lot of code like:

```fortran
if (my_rank == 0) write (*, *) 'something'
```
I found on the internet a brilliant idea - redirect output to `/dev/null`!
I made a couple of helper functions in `files.F90`, and now we can do this instead:

```fortran
# In init.F90
if (my_rank > 0) call stdout_to_devnull()

# Elsewhere
use mod_files, only: stdout
write (stdout, *) 'something'
```
I also wrote similar helpers for stderr. In this commit I converted mostly only write statements that mattered, so there are still a lot of `write (*,*)` which we can convert over time. I also converted a bunch of statements printing various errors to stderr.

Because of various module dependencies, I also needed move a couple functions into different modules, but in each of those the new placement actually made a lot more sense so I was quite happy to do it.

Because of this, I was able to get rid of global my_rank entirely,
in favour of using `get_mpi_rank()` from the new mpi module where needed.

### DIFF
--- a/src/en_restraint.F90
+++ b/src/en_restraint.F90
@@ -112,7 +112,7 @@ contains
                read (901, *, IOSTAT=ios) fxes(iat), fyes(iat), fzes(iat)
                if (ios /= 0) then
                   call fatal_error(__FILE__, __LINE__, &
-                     'Could not read gradients from file '//chforce_exc)
+                                   'Could not read gradients from file '//chforce_exc)
                end if
                ! Conversion to forces
                fxes(iat) = -fxes(iat)
--- a/src/nosehoover.F90
+++ b/src/nosehoover.F90
@@ -171,7 +171,7 @@ contains
    subroutine nhc_init()
       use mod_files, only: stdout
       if (imasst == 1) then
-          write (stdout, *) 'Initializing massive Nosé-Hoover Chain thermostat'
+         write (stdout, *) 'Initializing massive Nosé-Hoover Chain thermostat'
       else
          write (stdout, *) 'Initializing global Nosé-Hoover Chain thermostat'
       end if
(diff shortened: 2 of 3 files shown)


src/error.F90
```diff
@@ -56,6 +56,13 @@ subroutine fatal_error(filename, line_number, message)
       call error_method(filename, line_number, message=message)
    end subroutine fatal_error
 
+   subroutine not_compiled_with(feature)
```
danielhollas, line 59: Moved here from mod_utils

src/init.F90
```diff
@@ -968,46 +920,365 @@ subroutine print_read_error(chfile, chmsg, iost)
...
+      write (stdout, *) ' D. Hollas, J. Suchan, O. Svoboda, M. Oncak, P. Slavicek'
+      write (stdout, *) ' '
    end subroutine print_logo
 
 end subroutine init
 
+   ! Subroutine initialize_masses() populates the global am() array,
+   ! based on the atom names from names() array.
+   ! User can also specify non-standard isotopes/elements.
+   subroutine initialize_masses(masses, massnames, natom)
```
danielhollas, line 1002: Moved here from mod_system in `modules.F90`

danielhollas:
PR #100 merged! :tada: @suchanj thank you for all the reviews. :bow: I am basically done with the end-to-end tests, code coverage is now at respectable 87% percent. I have a few more refactors in mind.

- [ ] `force_abin` is a mess right now and can be simplified by splitting into smaller functions, which we can that unit test
- [ ] same goes for the main surface hopping routine and the main init, splitting into manageable chunks.
- [ ] remove the `itrj` index from all the SH arrays. This was never useful, and if we even need to run SH trajectories in parallel, it will be better done with MPI.

suchanj:
Hurray! 🥇Thanks @danielhollas for bringing in the best coding practices. They will definitely make ABIN relevant for the future.

PR #101, opened 2022-03-31
Fix the code coverage for mpi error paths

danielhollas:
Follow-up on #97. Do not call `MPI_Abort()` for TC-MPI interface, since it screws up code coverage for some reason (very strange :scream:), and it is not needed anyway since we only have one MPI process.

Add a test case for failing REMD.

Also minor cleanup in `remd_init()`.

### DIFF
--- a/src/mpi_wrapper.F90
+++ b/src/mpi_wrapper.F90
@@ -65,10 +65,17 @@ contains
 
       call MPI_Finalized(finalized, ierr)
 
-      if (error_code /= 0) then
-         call MPI_Abort(MPI_COMM_WORLD, error_code, ierr)
-      else if (.not. finalized) then
-         call MPI_Finalize(ierr)
+      if (.not. finalized) then
+         ! NOTE: For some reason, MPI_Abort does not play nicely with code coverage.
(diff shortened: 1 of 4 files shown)


src/mpi_wrapper.F90
```diff
@@ -65,10 +65,17 @@ subroutine finalize_mpi(error_code)
...
-      else if (.not. finalized) then
-         call MPI_Finalize(ierr)
+      if (.not. finalized) then
+         ! NOTE: For some reason, MPI_Abort does not play nicely with code coverage.
+         ! Therefore, for the case of just one MPI replica, e.g. for pot == _tera_,
+         ! we only call MPI_Finalize() and terminate by `stop error_code`.
+         ! For REMD, we need to ensure that all replicas are stopped,
+         ! so MPI_Abort is safer.
+         if (error_code /= 0 .and. get_mpi_size() > 1) then
+            call MPI_Abort(MPI_COMM_WORLD, error_code, ierr)
```
danielhollas, line 75: This line is in fact covered by the new REMD test case, but Codecov does not see it. I Guess that MPI_Abort is so aggressive that it somehow stop the code coverage generation prematurely.

PR #102, opened 2022-03-31
GHA: support Plumed 2.8

### LINKED ISSUE #83: Plumed 2.8.0 fails metadynamics unit test
https://github.com/PHOTOX/ABIN/runs/5409736811?check_suite_focus=true

Metadynamics forces produce non-negligible numerical difference in the unit test. We need to investigate if this is a possible bug in Plumed, or some change in interface or something else. There _were_ apparently some changes in the Fortran interface in 2.8.0, perhaps some of them were not backwards compatible.

```
ERROR STOP *** Encountered 1 or more failures/errors during testing. ***

Error termination. Backtrace:
Time:         0.032 seconds
  
Failure
 in: 
test_plumed_suite.test_force_metad
  Location: 
[test_plumed.pf:320]
fx(1) changed
AssertEqual failure:
      Expected: <0.99991611203271580>
        Actual: <0.99991594977761633>
    Difference: <-0.16225509946732330E-006> (greater than tolerance of 0.0000000000000000)
```

danielhollas:
Closes #83.

Plumed 2.8 fails one unit test and the METAD test because Plumed started using stretched gaussians, which leads to numerical differences.

See https://github.com/plumed/plumed2/issues/420 and https://github.com/plumed/plumed2/pull/757. As a quick fix for our test suite, we can set env variable PLUMED_DP2CUTOFF_NOSTRETCH before the linking step.

### DIFF
--- a/src/plumed.F90
+++ b/src/plumed.F90
@@ -35,16 +35,14 @@ module mod_plumed
 contains
 
 #ifdef USE_PLUMED
-   subroutine plumed_init(silent_output)
+   subroutine plumed_init()
       use mod_general, only: natom, irest, dt0, nrest
       use mod_const, only: ANG, AUTOFS, AMU, AVOGADRO, AUTOJ
       use mod_utils, only: c_string
       use mod_error, only: fatal_error
+      use mod_files, only: stdout
       !use mod_nhc, only: temp
       implicit none
-      ! Do not print any output (used in unit tests, unittest/test_plumed.pf)
-      logical, optional :: silent_output
-      logical :: silent
       integer, parameter :: PLUMED_RESTART = 1
       ! Conversion from ABIN to PLUMED units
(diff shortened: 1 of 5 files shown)


src/plumed.F90
```diff
@@ -62,22 +60,15 @@ subroutine plumed_init(silent_output)
...
       call plumed_f_gcreate()
 
       call plumed_f_gcmd(c_string("getApiVersion"), api_version)
 
-      if (.not. silent) then
-         write (*, '(a)') 'PLUMED is ON'
-         write (*, '(a)') 'PLUMED input file is '//trim(plumedfile)
-         write (*, '(a, i0)') 'PLUMED API version: ', api_version
-      end if
+      write (stdout, '(a)') 'PLUMED is ON'
```
danielhollas, line 69: So much nicer. :blush:

suchanj (approved):

PR #104, opened 2022-04-01
Use the correct decoherence correction by default

danielhollas:
Due to the error in the equation in the original paper, the decoherence correction was not implemented correctly
(never take anything for granted!).

This was already fixed in commit c7b342b9ad2, here we switch the use of correct formula by default.

To reproduce old results, add `correct_decoherence=.false` in the input file in namelist `sh`.

Thanks to Basile Curchod and Antonio Prlj for catching this!


danielhollas:
@suchanj I believe you did a bit of testing on this and found that it didn't have a huge effect, right?

Once I am done with all the refactor, I will test the SH code thoroughly again against the molecular Tully model molecules from Basile.

suchanj:
Yes, the limited testing proved no major deviations between new and old decoherence treatment.

danielhollas:
@suchanj FYI I am merging this for master for later testing, hopefully by the end of this week or next week.

PR #105, opened 2022-04-01
Fix PIMD energy for internal water potentials

### LINKED ISSUE #66: Wrong energy from water potentials for PIMD
Energy must be divided by number of beads, preferably in `forces.F90` (as opposed to inside the various force functions).
We will need to fix the PIMD test after the fix.

danielhollas:
Fixes #66


src/water_interface.cpp
```diff
@@ -92,14 +92,17 @@ extern"C" {
...
+      for (int iat = 0; iat < *natom; iat++) {
+         fx[iw * (*natom) + iat] = -grd[3*iat] * FAC;
+         fy[iw * (*natom) + iat] = -grd[1 + 3*iat] * FAC;
+         fz[iw * (*natom) + iat] = -grd[2 + 3*iat] * FAC;
       }
 
+   // nwalk loop
    }
 
+   *eclas /= *nwalk;
```
danielhollas, line 104: This is the fix

danielhollas:
I still need to update a couple of tests but otherwise ready for review.

suchanj (approved):

PR #106, opened 2022-04-01
Do not print file headers when restarting

danielhollas:
Aside from being nicer, this will also enable to more easily test that we get the same results when restarting.


suchanj:
Well I did not mind having the headers there to remind me restart was needed. However with properly functioning code they should not be needed so feel free to go ahead.

suchanj (approved):

danielhollas:
@suchanj good point, that was also a reason why I didn't mind them. This commit makes it easier to process the files, since one can just skip the first line, and don't worry about anything extra. But my main motivation here was for tests.

Specifically, I wanted to be able to easily test whether restarted simulation gives the same result as the continuous simulation. Now I can do something like this:

```
mv tests/CMD/intput.in2 tests/MORSE/intput.in
sed -i 's/irest=1/irest=0/' tests/MORSE/input.in && rm tests//restart.xyz.*.ref
make test TEST=MORSE
``` 
and see whether the test passes. Previously the extra header lines would mess that up.
I already previously noticed a small issue with SH restart, will open an issue and fix next week.


Nevertheless, with the refactor done here we can easily go back if we want.

Issue #107, opened 2022-04-03
Custom types for holding simulation state


danielhollas:
Maybe something like this

 ```fortran  
   type :: energies
      ! Ab initio electronic energy
      real(DP) :: e_pot
      ! Harmonic energy of the PI bead necklace
      real(DP) :: e_pi
   end type

   type :: forces
      ! Ab initio forces
      real(DP), allocatable, dimension(:, :) :: fxc, fyc, fzc
      ! Harmonic PI forces
      real(DP), allocatable, dimension(:, :) :: fxq, fyq, fzq
      ! Difference forces between full and reference potential
      real(DP), allocatable, dimension(:, :) :: fxdiff, fydiff, fzdiff
   end type

   type :: posmom
      ! Ab initio forces
      real(DP), allocatable, dimension(:, :) ::x, y, z
      real(DP), allocatable, dimension(:, :) ::px, py, pz
   end type
```

PR #108, opened 2022-04-03
Mdstep refactor

danielhollas:
Trying to move more stuff out of the big init subroutine into smaller chunks.

I also wanted to make a nice interface to various integrators.  I am hitting a problem here that various integrators have slightly different interface, e.g. some of them need extra force arrays. This problem in general and different ways to solve it are outlined here:
https://fortran-lang.org/learn/best_practices/type_casting

Even before this refactor this problem manifested for the MTS RESPA (`doublerespastep()`) integrator, where we have additional forces fx_diff, which we do not pass in as arguments but import them from global module. This is equivalent to  [approach no. 3 from the link above](https://fortran-lang.org/learn/best_practices/type_casting#private-module-variables), and I have utilized this approach for the other integrators as well so that I could make a minimal common interface (see top of mdstep module). Long terms it would be nice to define our own derive types that would hold the simulation state and that we could pass explicitly into all functions ([approach no. 2](https://fortran-lang.org/learn/best_practices/type_casting#general-structure)). I have outline how that could look in #107.

There are couple more unrelated refactors, mostly to pass data explicitly via subroutine arguments, instead of importing them from modules. This makes it more clear what depends on what.


src/forces.F90
```diff
@@ -276,77 +276,3 @@ subroutine force_quantum(fx, fy, fz, x, y, z, amg, quantum_energy)
    quantum_energy = equant
 end subroutine force_quantum
 
-! Based on:
```
danielhollas, line 279: Moved to mdstep.F90

src/transform.F90
```diff
@@ -206,15 +219,6 @@ subroutine initialize_pi_masses(amg, amt)
                ! write(*,*)'amg2',(amg(iat,iw)*2*sin(pi*(iw-1)/nwalk),iw=1,nwalk)
             end if
          end do
-
-      else if (inormalmodes == 1) then
```
danielhollas, line 210: This case is covered by default, see top of the function

src/abin.F90
```diff
@@ -163,21 +163,8 @@ program abin
...
-            ! include entire Ehrenfest step, in first step, we start from pure initial state so at first step we dont
-            ! need NAMCE and take forces just as a grad E
-         case (3)
-            call respashake(x, y, z, px, py, pz, amt, amg, dt, equant, eclas, fxc, fyc, fzc, fxq, fyq, fzq)
-         case (4)
-            call doublerespastep(x, y, z, px, py, pz, amt, amg, dt, equant, eclas, fxc, fyc, fzc, fxq, fyq, fzq)
-         end select
-
-         sim_time = sim_time + dt
+         ! PROPAGATE through one time step
```
danielhollas, line 166: Here's the nice part.

src/mdstep.F90
```diff
@@ -4,14 +4,96 @@
...
-   public :: verletstep, respastep, respashake, doublerespastep
+   public :: initialize_integrator
+
+   ! Denotes number of internal steps in multiple-time-step RESPA algorithms
+   ! nabin - internal steps in PIMD integration,
+   integer, public :: nabin = 50
+   ! nstep_ref - internal steps with reference potential pot_ref
+   integer, public :: nstep_ref = 1
+
+   abstract interface
```
danielhollas, line 20: Gettting a bit fancy here, here's a nice explanation:
https://fortran-lang.org/learn/best_practices/callbacks

src/init.F90
```diff
@@ -261,20 +265,19 @@ subroutine init(dt)
...
    names = ''
    attypes = ''
    massnames = ''
-   masses = -1.0D0
 
-   allocate (ishake1(natom * 3 - 6))
-   allocate (ishake2(natom * 3 - 6))
-   ishake1 = 0
-   ishake2 = 0
+   allocate (ishake1(natom * 3 - 6), source=0)
```
danielhollas, line 279: New trick, allocation and initialization in one statement.

src/abin.F90
```diff
@@ -276,24 +263,9 @@ program abin
...
-   end if
-
-   write (stdout, *) ''
-   write (stdout, '(A)') 'Job finished successfully!'
-
+   call print_footer(time_start)
    call finish(0)
 
-   call cpu_time(total_cpu_time)
-   write (stdout, '(A)') 'Total cpu time [s] (does not include ab initio calculations)'
```
danielhollas, line 288: I removed the CPU time printing, because it's not indicative of anything as it does not account for external program execution.

src/en_restraint.F90
```diff
@@ -42,15 +42,13 @@ end subroutine en_rest_init
...
       use mod_general, only: natom, nwalk, dt0 ! dt0 is the time step
-      use mod_const, only: AMU
       use mod_system, only: am
       use mod_sh, only: en_array
       use mod_terampi_sh, only: force_terash
       real(DP), intent(inout) :: x(:, :), y(:, :), z(:, :)
       ! TODO: Eclas is not modified in this routine, but probably should be
       real(DP), intent(inout) :: eclas
       real(DP), intent(inout) :: px(:, :), py(:, :), pz(:, :)
-      ! DH: I am surprised this works, since natom is not a constant
```
suchanj, line 53: Btw do you have the answer why this works? Maybe the variable is initialized once the subroutine is called and at that time the natom is known?

src/mdstep.F90
suchanj: That's a cool trick. 'Welcome pointers my old friend.'

suchanj (approved):

src/mdstep.F90
danielhollas: Yeah. The good thing here is that the logic for selecting the integrator is localized in one place, whereas before it was both in init.F90 and abin.F90.

src/en_restraint.F90
danielhollas: Exactly. These are so called [automatic arrays](https://stackoverflow.com/questions/31433121/difference-between-local-allocatable-and-automatic-arrays) that get allocated on the stack when the function is called.
Instead of relying on the global `natom` variable, a slightly better way would be to use the `size()` intrinsic to get the size of the input arrays, like this:
```fortran
real(DP) :: a(size(x, 1), size(x, 2))
```

danielhollas: The disadvantage in comparision to allocatable arrays that are allocated on the heap is that stack size is usually limited. So if we were simulating big system, we might get a segfault / stack overflow. So in general manually allocating arrays of size (natom, nwalk) is probably a bit better for the future. 

On Unix, the stack size can be increased by
```bash
ulimit -s unlimited
```

danielhollas: Actually, reading that SO link I posted above, looks like whether automatic arrays are placed on the stack or on the heap depends on the compiler and compiler options. So if this ever became an issue in the future, tweaking compiler options would solve it.

PR #109, opened 2022-04-05
Refactor force_abin

danielhollas:
I have split the rather complicated force_abin() functions into smaller subroutines and put them all inside a module. I was them able to reused them in the oniom subroutine. 

Also some small cleanup in restart functions, and some further conversions from abinerror to fatal_error.

I also created a helper util function `append_rank`, for modifying the file names for REMD.


danielhollas:
@suchanj I am still tweaking unit tests, but otherwise this is good to review. The diff for force_abin.F90 is messy since I moved a lot of code, so it's perhaps better to checkout the branch and see the file as a whole. Thanks!

suchanj (approved):

PR #110, opened 2022-04-06
Robust and reproducible PRNG initialization for REMD

danielhollas:
This effort started because I wanted to get rid of Gfortran-specific routines that we used to generate additional random seeds for REMD so that we can compile REMD code with ifort.

My original plan was to use the `random_number()` subroutine that is defined by the Fortran standard. This is now implemented in the new `random_ints()` subroutine, and I have refactored the `utils/abin_randomint.f90` to use it.

Unfortunately, using it to generate seeds for REMD is problematic for two reasons:
1. Results would depend on compiler and compiler versions. Not ideal for reproducibility, and not usable for tests.
2. Seeding `random_number()` is not as easy, and requires a compiler-dependent array of integers, so we are back to square one.

In the end, I have used a simple prng that I found on the Gfortran manual, to both generate seeds for REMD, and to deterministically initialize `random_number()` (function `lcg()`)

As a new functionality, I have added automatic determination of initial random seed from `/dev/urandom` or current date and time. (more info in code comments). I am not sure whether users are always careful with changing their seeds in the input file, so I made this a new default and removed `irandom` parameter from sample input files. @suchanj let me know your thoughts on this.

The new code is in a separate module `mod_prng_init` and contains a lot of comments explaining the situation.

TODO: 
- [x] Add unit tests
- [x]  Test all new functions, do they produce uniform prng distributions?
- [x]  Refactor and test saving/loading prng state
- [x] add taking initial seed from `/dev/urandom`
- [x] convert abinerror to fatal_error()
- [x] remove dead code (unrolling optimization)


suchanj:
The automatic determination of initial seed is a wonderful idea, I don't think lot of people took good care of that.

suchanj (approved):

PR #111, opened 2022-04-18
Include ifort in test suite

### LINKED ISSUE #27: Stabilize tests

### LINKED ISSUE #73: Include IntelONE API compiler in Github Actions
For reference, see
https://github.com/grimme-lab/crest/blob/master/.github/workflows/build.yml

Depends on #27

danielhollas:
The goal here is for the `ifort` intel compiler to pass the test suite without reducing the threshold in `numdiff.py`.

Turns out most of the issues came from `write (*,*)` statements, where ifort compiler would sometimes round the last digits of real numbers differently. Specifying format explicitly as `ES24.16E3` (i.e. full double precision) fixes that.

Since the latest versions of the Intel compilers are now open source (Intel OneAPI), we can add it to the GHA builds! :tada: 

Closes #27. Closes #73.

Blocked on #110 , which fixes REMD tests for ifort.


src/files.F90
```diff
@@ -259,7 +259,7 @@ subroutine print_file_headers()
       end if
       headers(UCHARGES) = '# Atomic charges from current electronic state'
       headers(UDOTPRODCI) = '# Time  cidotprod1  cidotprod2 ... '
-      headers(UTDIP) = '# Time  st  tdip_tot.1 tdip_tot.2 ... tdip_x.1 tdip_y.1 tdip_z.1 tdip_x.2 tdip_y.2 tdip_z.2.'
```
danielhollas, line 262: Fortran would break up this long line into two :-(

src/landau_zener.F90
```diff
@@ -507,7 +507,7 @@ subroutine lz_restout(fileunit)
 
       write (fileunit, *) istate_lz
       do ist = 1, nstate_lz
-         write (fileunit, *) en_array_lz(ist, 1), en_array_lz(ist, 2), en_array_lz(ist, 3)
+         write (fileunit, '(3ES25.16E3)') en_array_lz(ist, 1), en_array_lz(ist, 2), en_array_lz(ist, 3)
```
danielhollas, line 510: One of the test was failing with ifort because of this.

tests/CMD/energies.dat.ref
```diff
@@ -1,21 +1,3 @@
...
-           5.81    0.3384438868E-02    0.2089517656E-02    0.5473956524E-02    0.2851843585E-02
-           6.29    0.3595636485E-02    0.2614491942E-02    0.6210128427E-02    0.3110173188E-02
-           6.77    0.3197088006E-02    0.4233089278E-02    0.7430177284E-02    0.3418744909E-02
-           7.26    0.2724805517E-02    0.4660775300E-02    0.7385580818E-02    0.3683200636E-02
-           7.74    0.2368842821E-02    0.4181320077E-02    0.6550162898E-02    0.3862385778E-02
-           8.22    0.2005745173E-02    0.4334899133E-02    0.6340644305E-02    0.4008165691E-02
-           8.71    0.1741443137E-02    0.4595150928E-02    0.6336594065E-02    0.4137522823E-02
-           9.19    0.1721145984E-02    0.4220770907E-02    0.5941916891E-02    0.4232490932E-02
-           9.68    0.1790041359E-02    0.3640889399E-02    0.5430930758E-02    0.4292412923E-02
+           0.97    0.6284700637E-03    0.7797219590E-03    0.1408192023E-02    0.1740580924E-02
```
danielhollas, line 3: TODO: This line after restart should be the same as before. It seems that Nose-Hoover does not give precisely the same results after restarts.

tests/numdiff.py
```diff
@@ -114,7 +114,7 @@ def parse_diff(fname, absolute_tolerance):
 
 
 if __name__ == '__main__':
-   absolute_tolerance = 1e-15
+   absolute_tolerance = 2e-15
```
danielhollas, line 117: Sometimes the last digit differs.

PR #118, opened 2022-04-21
Fix intent in lz_restin

### LINKED ISSUE #74: Invalid intent for input parameters in `lz_restin`
I got tipped off by a compiler warning. It looks like the position and velocity arrays passed into `lz_restin` should be `intent(in)` instead of `intent(out)`. 

https://github.com/PHOTOX/ABIN/blob/4ca60042ab0f922f0202e14b42615761a2acd56b/src/landau_zener.F90#L515

Note that this is actually a bug, since variables declared as `intent(out)` are not guaranteed to have the values that you passed in, so you might be assigning random numbers here.

https://github.com/PHOTOX/ABIN/blob/4ca60042ab0f922f0202e14b42615761a2acd56b/src/landau_zener.F90#L542

I don't understand why the compiler proceeds without error here, supposedly the purpose of the `intent` attribute is to prevent such mistakes. :frowning:

danielhollas:
Fixes #74

I am surprised that this has worked, and also that older compiler versions didn't catch this mistake.

@suchanj please take a look as this is your code.

### DIFF
--- a/src/landau_zener.F90
+++ b/src/landau_zener.F90
@@ -512,8 +512,8 @@ contains
 
    subroutine lz_restin(fileunit, x, y, z, vx, vy, vz)
       integer, intent(in) :: fileunit
-      real(DP), intent(out) :: x(:, :), y(:, :), z(:, :)
-      real(DP), intent(out) :: vx(:, :), vy(:, :), vz(:, :)
+      real(DP), intent(in) :: x(:, :), y(:, :), z(:, :)
+      real(DP), intent(in) :: vx(:, :), vy(:, :), vz(:, :)
       integer :: ist
 
       read (fileunit, *) istate_lz


danielhollas:
One remaining gfortran warning is

```sh
landau_zener.F90:450:28:

  450 |    subroutine lz_getgraddiff(grad_diff, ist, ist1, x, y, z, vx, vy, vz, fxc, fyc, fzc)
      |                            ^
Warning: ‘lz_getgraddiff’ defined but not used [-Wunused-function]
```
@suchanj we can live with this, but let me know in case it is okay to remove this function.

suchanj:
Wow, yeah that should not have worked at all. (LZ restart is covered by tests so it somehow did.)
I would actually keep that function, Jirka Janoš is now looking into LZ and there is a plan to base the PES on the force integration so it will at least help as a reference. Sorry for keeping the code bit messy.

suchanj (approved):

suchanj:
Please, wait a sec with merging ...

suchanj:
Thanks, all clear!

PR #119, opened 2022-04-21
convert all system() calls to execute_command_line()

### LINKED ISSUE #116: Convert system calls to `execute_command_line`
system() is a GNU intrinsic, although it does appear to work for ifort. `execute_command_line` subroutine comes from Fortran 2008 standard and is hopefully supported by recent compilers.

https://gcc.gnu.org/onlinedocs/gfortran/EXECUTE_005fCOMMAND_005fLINE.html#EXECUTE_005fCOMMAND_005fLINE

danielhollas:
Closes #116. I was able to get rid of a couple of system calls by introducing utility functions for deleting and renaming files.

I also realized that the error handling inside the parallelized OpenMP do loop in force_abin is not according to the OpenMP standard so needed to do a little refactor.


src/force_abin.F90
```diff
@@ -220,20 +237,27 @@ subroutine force_abin(x, y, z, fx, fy, fz, eclas, chpot, walkmax)
...
       ! e.g. ./G09/r.g09
       shellscript = get_shellscript(chpot)
+      abort = .false.
 
-!$OMP PARALLEL
-!$OMP DO PRIVATE(engrad_unit, energy, chgeom, chforce)
+!$OMP PARALLEL DO PRIVATE(engrad_unit, energy, chgeom, chforce)
       do iw = 1, walkmax
 
+      ! If one PI bead external calculation fails, we need to stop as soon as possible,
```
danielhollas, line 254: This is annoying, but thanks to the earlier refactor, the workaround is hopefully not too complex.

src/landau_zener.F90
```diff
@@ -447,6 +445,7 @@ subroutine lz_rewind(en_array_lz)
       en_array_lz(:, 2) = en_array_lz(:, 3)
    end subroutine lz_rewind
 
+#if 0
```
danielhollas, line 448: Commenting this code out of compilation for now to make clear it is not used and to remove the compiler warning.

src/abin.F90
```diff
@@ -2,7 +2,7 @@
 !  ABIN: Multipurpose ab initio MD program.
 !  The potential is calculated on-the-fly by an external program.
 !------------------------------------------------------------------
-!  Copyright (C) 2014             D.Hollas, M.Oncak, O.Svoboda and P.Slavicek
+!  Copyright (C) 2014    D.Hollas, J.Suchan, M.Oncak, O.Svoboda and P.Slavicek
```
suchanj, line 5: 😱

suchanj (approved):

src/en_restraint.F90
```diff
@@ -161,7 +175,7 @@ subroutine energy_restraint(x, y, z, px, py, pz, eclas)
    end subroutine energy_restraint
 
    subroutine en_rest_finalize()
-      deallocate (fxr, fyr, fzr)
+      if (allocated(fxr)) deallocate (fxr, fyr, fzr)
```
danielhollas, line 178: @suchanj just FYI in finalization routines one should always check whether things are allocated before deallocating them since ABIN can fail early before the inititialization routine is called.

PR #121, opened 2022-04-25
Refactor init

### LINKED ISSUE #57: Use `open(newunit=unit..` instead of random integers
This is a Fortran 2008 that should be available in all modern compilers. I already successfully use it in `unittests/test_plumed.pf`.
This applies to many random places around the codebase, but also in `mod_files` routines.
As part of that, we should move `mod_files` from `modules.F90` to its own file.

### LINKED ISSUE #82: Convert all calls to abinerror() to fatal_error()

danielhollas:
I've been cleaning up the init() subroutine in previous PRs, this is the last bit. The only remaining refactor and improved testing concerns the surface hopping routines.

Also in this PR I've removed the last calls to `abinerror()`. :tada: 

Closes #82. Closes #57.


src/modules.F90
```diff
@@ -46,7 +46,7 @@ module mod_general
    integer :: icv = 0, anal_ext = 0, idebug = 0
    integer :: ihess
    ! Number of atoms, taken from XYZ geometry
-   integer :: natom = 0
+   integer, protected :: natom = 0
```
danielhollas, line 49: In general, if we need to have global variables, at least they should be read-only.

src/plumed.F90
```diff
@@ -35,14 +35,14 @@ module mod_plumed
 contains
 
 #ifdef USE_PLUMED
-   subroutine plumed_init()
-      use mod_general, only: natom, irest, dt0, nrest
+   subroutine plumed_init(natom, irest, dt0, nrest)
```
danielhollas, line 38: Better be explicit about what parameters are neede for PLUMED initialization (and in general).

src/plumed.F90
```diff
@@ -196,7 +196,12 @@ end subroutine force_plumed
    ! Stubbed subroutines for compilation without PLUMED
    ! We use this approach to avoid needing to have '#ifdef USE_PLUMED'
    ! elsewhere in the codebase
-   subroutine plumed_init()
+   subroutine plumed_init(natom, irest, dt0, nrest)
+      integer, intent(in) :: natom, irest, nrest
+      real(DP), intent(inout) :: dt0
+      dt0 = natom
+      dt0 = irest
```
suchanj, line 203: Why is this assignment needed?

suchanj (approved):

src/plumed.F90
danielhollas: Just to squash compiler warnings. :-(

suchanj: Ok, I wouldn't mind comment in the code but its alright.

src/init.F90
```diff
@@ -117,7 +122,7 @@ subroutine init(dt)
       !
       ! All namelists need to be in a single input file, and the code
       ! in this subroutine must ensure that the namelists can be in any order.
-      namelist /general/ natom, pot, ipimd, mdtype, istage, inormalmodes, nwalk, nstep, icv, ihess, imini, nproc, iqmmm, &
```
danielhollas, line 120: @suchanj I forgot to mention that there is one breaking change in this PR; I removed the `natom` keyword from the input. It has been deprecated and ignored for a long time so I think it is okay to remove it. I think it is fine to require users to delete it from their input files if they still have it, but LMK if you disagree.

PR #123, opened 2022-04-26
Fix reading velocities from a file in r.terabin

danielhollas:


danielhollas:
Ugh, sorry, wrong branch, will open another...

PR #125, opened 2022-04-28
SH: Remove NSTMAX

danielhollas:
Allocate all SH arrays dynamically based on the number of states.


src/landau_zener.F90
```diff
@@ -13,30 +13,30 @@ module mod_lz
...
    integer :: nstate_lz, nsinglet_lz = 0, ntriplet_lz = 0
 
    ! Module variables
    integer :: istate_lz
-   real(DP), allocatable :: en_array_lz(:, :), en_array_lz_backup(:, :)
-   real(DP), allocatable :: fx_old(:, :), fy_old(:, :), fz_old(:, :)
-   real(DP), allocatable :: px_temp(:, :), py_temp(:, :), pz_temp(:, :)
-   real(DP), allocatable :: x_prev(:, :), y_prev(:, :), z_prev(:, :), &
-                            vx_prev(:, :), vy_prev(:, :), vz_prev(:, :)
+   ! TODO: Is this array needed? Are we not computing gradients for only one state?
```
danielhollas, line 33: @suchanj I am wondering, in LZ do you  ever compute gradients for more than one state?

src/modules.F90
```diff
@@ -3,17 +3,6 @@
 ! We are using modules to initialize some variables and
 ! for passing global variables to different subroutines.
 
-! mod_array_size contains various array limits.
-! Modify here if you need larger arrays.
-! Most of the other arrays are allocated dynamically.
-module mod_array_size
```
danielhollas, line 9: :tada:

src/landau_zener.F90
suchanj: No, not right now, however we might if J. Janoš implements his PES by gradient integration method.

src/surfacehop.F90
```diff
@@ -31,16 +30,20 @@ module mod_sh
    public :: nacx, nacy, nacz
 
    ! Initial electronic state
-   integer :: istate_init = 1
+   integer :: istate_init = -1
```
danielhollas, line 33: I have changed the default to invalid number so that users don't forget to set initial state.

src/landau_zener.F90
danielhollas: Okay, great, I'll remove this comment then.

suchanj (approved):

src/landau_zener.F90
```diff
@@ -77,6 +77,8 @@ subroutine lz_init()
       !Initial state
       istate_lz = initstate_lz
 
+      allocate (tocalc_lz(nstate_lz))
```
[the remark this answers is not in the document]
danielhollas: Thanks good catch.  While we're here, could you check in lz_finalize that at least one of the arrays is indeed allocated? Otherwise if lz_fibalize is called before lz_init we'd get runtime error.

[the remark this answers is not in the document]
suchanj: Yep, good idea, done.

PR #126, opened 2022-04-28
SH: Cleanup the main SH routine

danielhollas:
This is mostly just moving code out of the main SH loop into separate subroutines. 

One minor breaking change is that I have renamed the files in which we store geometries where the hop occured. 
Instead of `geom.<oldstate>.<newstate>.timestep>` we have `hopgeom.oldstate.newstate.timestep.xyz`

Changes in test files are due to random number state, since we are now calculateing hopping probabilities even if hopping is forbidden via `nohop=1`. We now print a warning if a hop is ignored because of this setting.


src/surfacehop.F90
```diff
@@ -668,76 +643,27 @@ subroutine surfacehop(x, y, z, vx, vy, vz, vx_old, vy_old, vz_old, dt, eclas)
...
-            ! Auxiliary calculations of probabilities on a number line
-            prob = 0.0D0
-            if (ist == 1) then
-               ! If we are in the ground state, we cannot jump into the ground state :-)
-               prob(1) = 0.0D0
-            else
-               ! Probability of jumping from the current state to the ground state
-               prob(1) = t(ist, 1)
-            end if
+         call random_hop(transmat=t, old_state=ist, new_state=ihop)
```
danielhollas, line 646: It might be worth checking out these changes locally to see that the main SH loop is now much cleaner. :blush:

suchanj (approved):

PR #131, opened 2022-05-05
Fix MOLPRO-SH interface

danielhollas:
When I refactored force_abin in #109, I deleted a bit of obscure logic in `read_forces()` that allowed skipping forces. 
Turns out this logic was needed for SH, since in `sh_init` we were calling the interface and requesting only energies without forces.  Hence, that change broke the MOLPRO-SH interface. I could reintroduce the hack or solve it in some other way, but it felt easier for now to simply calculate the forces, even thoug hwe through them out in `sh_init`.

I have also added a workaround for a compiler bug that is specific for the Gfortran 7.3 (which happens to be installed on PHOTOX clusters unfortunately). The workaround is neede for only in the error paths in the file interface in case energies or forces could not be read. Before the workaround, the `inquire` statement that we are using to get the filename from the file unit would result in an ugly and confusing runtime error.


src/surfacehop.F90
```diff
@@ -116,13 +116,14 @@ subroutine sh_init(x, y, z, vx, vy, vz)
          istate = istate_init
       end if
 
-      ! computing only energies, used for subsequent
-      ! determination of TOCALC according to deltaE threshold
       allocate (tocalc(nstate, nstate))
       tocalc = 0
+      tocalc(istate, istate) = 1
```
danielhollas, line 121: This is the fix.

tests/SH_BUTCHER/MOLPRO/r.molpro
```diff
@@ -2,10 +2,22 @@
 cd MOLPRO
 timestep=$1
 ibead=$2
+natom=$(wc -l < ../geom.dat.$ibead)
```
danielhollas, line 5: I have improved this test case so that it would have caught the bug.

unit_tests/test_interface.pf
```diff
@@ -94,14 +93,19 @@ contains
       integer :: u
 
       call del_file(fname)
-      ! Working around a bug in Gfortran 7.3.0, where the inquire statement
```
danielhollas, line 97: This is now solved by the conditional compilation in force_abin.

suchanj (approved):

PR #136, opened 2022-06-13
SH: Fix edge condition in try_hop_nacme_rescale()

### LINKED ISSUE #135: Guard against division by zero `try_hop_rescale_nacme`
An interesting edge case was found when running SH with small `deltaE` threshold. In that case, a hop between states could happen right between the steps where the energy separation exceeds the threshold, and NACME become zero. The leads to division by zero since we are trying to scale velocities along the zero NAC vector. 

https://github.com/PHOTOX/ABIN/blob/63d62affd97de49371df4e72f371a26074105060/src/surfacehop.F90#L931

One solution would be to pass in the interpolated NACM, which should never be zero, since otherwise the hopping probability is also zero. However, that is slightly inconsistent since we're rescaling velocities at the end of the time step, not interpolated velocities. 

Alternatively, we could simply guard against this condition and call isotropic rescaling when this happens. 

Neither of these fixes is very satisfactory. The fundamental issue is that we're trying to hop every substep, but actually then the rescaling a recalculation happens at the end of the big step. To make things really consistent I think we should try to hop only once per big time step and accumulate hopping probability along the small time step. I know that both approaches are discussed in the literature, but I don't know what is the current consensus. We should check what SHARC/Newton-X does, and perhaps other SH packages.

danielhollas:
Closes #135, see that issue for details.

I will add a unit test for this in another PR.

### DIFF
--- a/src/surfacehop.F90
+++ b/src/surfacehop.F90
@@ -892,8 +892,10 @@ contains
       integer :: iat, iw
 
       iw = 1
-      ! Checking for frustrated hop
+      write (stdout, '(A,I0,A,I0)') 'Trying to hop from state ', instate, ' to state ', outstate
+      write (stdout, *) "Checking kinetic energy in the direction of NACME vector"
 
+      ! Checking for frustrated hop
       a_temp = 0.D0
       b_temp = 0.D0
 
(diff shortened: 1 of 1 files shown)


src/surfacehop.F90
```diff
@@ -910,16 +912,25 @@ subroutine try_hop_nacme_rescale(vx, vy, vz, instate, outstate, eclas)
       a_temp = 0.5D0 * a_temp
       c_temp = b_temp**2 + 4 * a_temp * (en_array(instate) - en_array(outstate))
 
+      if (a_temp <= 0.0D0 ) then
```
JanosJiri, line 915: Can this actually happen to have zero NACME when you just hopped? You need non-zero NACME for hopping.

src/surfacehop.F90
```diff
@@ -910,16 +912,25 @@ subroutine try_hop_nacme_rescale(vx, vy, vz, instate, outstate, eclas)
...
+         write (stdout, *) 'WARNING: NACME vector is zero, using isotropic velocity rescaling'
```
JanosJiri, line 916: I would rather write "rescaling along the velocity vector" instead of "isotropic". But its just a minor comment.

JanosJiri (approved):
I have just two minor comments but otherwise I see no problem.

PR #138, opened 2022-07-19
xTB interface

juraskov:
Simple interface for xTB v. 6.2

### DIFF
--- a/interfaces/XTB/r.xtb
+++ b/interfaces/XTB/r.xtb
@@ -1,6 +1,7 @@
 #!/bin/bash
 cd $(dirname $0)
-source ../SetEnvironment.sh XTB
+#source ../SetEnvironment.sh XTB
+XTBEXE=xtb
 
 timestep=$1
 ibead=$2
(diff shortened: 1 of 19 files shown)


interfaces/XTB/r.xtb
```diff
@@ -0,0 +1,61 @@
...
+if [[ $? -eq 0 ]];then
+   cp $input.out $input.out.old
+else
+   echo "WARNING from r.xtb: XTB calculation probably failed."
+   echo "See $input.out.error" 
+   cp $input.out $input.out.error
+fi
+
+### EXTRACTING ENERGY AND FORCES
+#Archives all energies and forces, possible to comment
```
danielhollas, line 52: Please remove the archiving, since energies are printed by ABIN anyway, and one can also print forces with `nwritef=1`

interfaces/XTB/r.xtb
```diff
@@ -0,0 +1,61 @@
...
+#Charge can be specified by changing value in a file .CHRG
+
+echo "0" > .CHRG
+
+cat > $input << EOF
+\$periodic 3
+\$cell angs
+    18.428    18.428    18.428  90  90  90
+\$coord angs
+
```
danielhollas, line 30: Are these two extra lines needed?

danielhollas (commented):
Awesome, thank you :heart: Just a couple of minor comments if you could address them. :bow:

danielhollas:
@juraskov Perfetto. :tada: :ok_hand: Thank you very much.

I did a bit of cleaning in other BASH interfaces and killed all useless cats. :drop_of_blood: I have also in the end simplified the XTB interface a bit more, I got rid of the input file altogether, and instead the geometry is passed in via standard XYZ file, and all the parameters are passed via command line arguments. I have also added number of processors and unpaired electrons as parameters to make this more general. I tested energy conservation for GFN1-xTB. Interestingly, when I tried GFN0-xTB for this particular molecule, it crashed, not sure why, and don't have time to dig in further. In any case, this will likely be very useful.

CC @suchanj

danielhollas (approved):

PR #145, opened 2022-09-26
Protocol buffer (TCPB) interface to TeraChem

### LINKED ISSUE #114: Migrate from Codecov BASH uploader
Codecov bash uploader is being deprecated
https://docs.codecov.com/docs/about-the-codecov-bash-uploader

New uploader:
https://docs.codecov.com/docs/codecov-uploader#introduction

Usage on GHA
https://github.com/marketplace/actions/codecov

### LINKED ISSUE #144: Implement TeraChem TCPB interface
The MPI interface for ground state MD has been removed in the recent TC version so we need to switch to an interface based on Google's protobuf. Fortunately, there's a tcpb-cpp library that despite its name also contains Fortran wrappers so it shouldn't hopefully be too hard to use. :crossed_fingers: 

https://github.com/mtzgroup/tcpb-cpp/tree/master/examples/api/fortran

### LINKED ISSUE #151: Codecov uploads failing for gcc9 and gcc10
This is affecting the TCPB build, which runs on Ubuntu 20.04 with GCC-9. GCC-7 on Ubuntu 18.04 works.

https://app.codecov.io/gh/PHOTOX/ABIN/commit/dc0ac9b6fe6eae19fd041a6a278e5b8ebcac1fec

image

Probably the same issue that I already reported back when we were using the BASH uploader. The solution might be the same, but need to figure out how to use it with the codecov-action on GHA.

https://community.codecov.com/t/intermittent-timeout-contacting-codecov-io-from-github-actions/1690/28

https://community.codecov.com/t/gcov-coverage-reports-were-not-found-when-migrating-from-codecov-codecov-action-v1-to-codecov-codecov-action-v2/3273

https://community.codecov.com/t/gcov-coverage-reports-were-not-found-when-migrating-from-codecov-codecov-action-v1-to-codecov-codecov-action-v2/3273/11

https://github.com/codecov/codecov-action
https://github.com/marketplace/actions/codecov

danielhollas:
In the recent TeraChem dev version they removed the Amber MPI interface that we were using for ground state MD. (FMS/MPI that we use for surface hopping is still there). 

As a replacement, here I implement the new TCPB interface which is based on communication via plain Unix socket and data are serialized using Google's [Protocol Buffers protocol](https://developers.google.com/protocol-buffers).
Fortunately, we don't need to deal with the details here thanks to the [tcpb-cpp](https://github.com/mtzgroup/tcpb-cpp), which offers a thin Fortran API. 

I didn't test this too extensively, but I did run a couple of 1 ps NVE dynamics and things seems to be stable and energy is conserved. Note, this will not work on PHOTOX clusters since TC was not compiled with this functionality.

I did not write a full test for this yet, since we would again need to mock the TC server, and I am not yet familiar how the TCPB interface works on TeraChem side. Will do it eventually, but need to get this out so we can compile on our clusters.
I did however add a TCPB build to GitHub CI, so we're at least testing the ABIN buillds with this functionality, and goes through some initialization (until the TC connection step, which inevitably fails).

QM/MM is not implemented yet, but shouldn't in principle hard to add once somebody needs it.
On the other hand, it seems that it is currently [not possible to connect to multiple TC servers](https://github.com/mtzgroup/tcpb-cpp_dev/issues/10) so REMD is not supported.

Closes #144 Closes #151
~~NOTE: Codecov is acting up again and does not pick up the TCPB build for some reason.~~


danielhollas:
I am essentially done with the code, what remains is to add a test (if it is not too much work) and documentation. 

The latest test failures are due to Codecov upload.

danielhollas:
@suchanj :wave: Take a look when you have a bit of time, but no rush. 

This is not something that you can test on PHOTOX clusters unfortunately, but it would be great to let me know if this all makes sense at least in principle.

danielhollas:
@suchanj are you okay with merging this?

suchanj:
Oh, I'm so sorry, must have missed the email! This explains why ground state MD does not run on krypton with current dev TC version.
Yes lets go ahead and merge it. The code structure makes sense. I guess it could be easily expanded if we would need to transition from FMS/MPI to PB completely.

suchanj (approved):

danielhollas:
> Oh, I'm so sorry, must have missed the email! This explains why ground state MD does not run on krypton with current dev TC version.

@suchanj  interesting, did you recompile TC recently? If so you could actually test this.

PR #150, opened 2022-11-02
LZ with TC fixes #142

### LINKED ISSUE #142: TC with LZ broken
sh_init routine interferes and terminates the calculation

suchanj:
Just a quick fix to make LZ work with TeraChem. 
Sadly, this is again temporary solution. We would need to separate MPI interface from SH to make it usable by other routines independently.
I also tried to write TERAMPI test, however, I would need to supply actual gradients and energies of multiple states which would be quite demanding given the current state. I would maybe wait until the aforementioned changes are done. 

Closes #142


suchanj:
Note: Do not rename PR, codecov will fail

src/landau_zener.F90
```diff
@@ -100,12 +100,27 @@ subroutine lz_init()
       ! TERA-MPI parameters
       if (pot == '_tera_') then
          nstate = nstate_lz
-         call set_current_state(initstate_lz)
```
danielhollas, line 103: I would propose to call `lz_init_terash` from here, instead of `abin.F90`. Makes more sense to me, especially since there is already is special code here. Also please pass the variable `pot` explicitly as a function parameter instead of importing it from module. It is better to be explicit that this is needed in this function.

danielhollas (changes requested):
Thanks @suchanj! I am glad that at least we're not calling the `sh_init` function when running LZ.

Please also cleanup the `sh_init` function, there's an extra LZ-related condition based on ipimd that is not needed anymore after these changes.

Regarding test, I think you should be able to copy and modify the existing `TERAPI-SH-S0` test. It's is not a full test, as it is essentially an MD on the ground state, but should serve as a basic [smoke test](https://en.wikipedia.org/wiki/Smoke_testing_(software)) nonetheless.

danielhollas:
Ah, I guess I know what you mean. Sometimes Codecov upload fails, but usually you can just re-run the job to fix it (You can trigger the re-run in the Github Actions page)

suchanj:
So I performed the cleanup and added the TERAPI-LZ test. I extended the tc_mpi_api with fms population reading energies and gradients from file. It is far from being perfect but could be used and improved for future tests. Also coverage :arrow_up:

src/landau_zener.F90
```diff
@@ -19,7 +19,7 @@ module mod_lz
    use mod_sh_integ, only: nstate
    implicit none
    private
-   public :: lz_init, lz_hop, lz_rewind, lz_restin, lz_restout, lz_finalize !Routines
+   public :: lz_init, lz_init_terash, lz_hop, lz_rewind, lz_restin, lz_restout, lz_finalize !Routines
```
danielhollas, line 22: This doesn't have to be public anymore.

src/landau_zener.F90
```diff
@@ -69,8 +69,9 @@ subroutine check_lz_parameters()
    end subroutine check_lz_parameters
 
    !Initialization
-   subroutine lz_init()
+   subroutine lz_init(pot)
```
danielhollas, line 72: Thanks for making it a paramater. You might maybe remove the import of `pot` from the top of the module, line 15. Also it seems that on the same line `irest` and `nwalk` are imported but not used.

danielhollas (approved):
@suchanj Amazing, thanks so much for the test.

Just a couple of final nitpicks but otherwise good to go.

PR #153, opened 2022-12-01
BAGEL interface for XMS and MS-CASPT2 in FSSH and LZSH

JanosJiri:
The interface for BAGEL and ABIN to perform FSSH and LZSH simulations. It is set to do XMS-CASPT2 by default but input can be modified for MS-CASPT2 and CASSCF. It was tested on cyclopropanone molecule and benchmarked to MOLPRO. Credits to Tomas Jira and Jiri Suchan.


interfaces/BAGEL-LZ/r.bagel
```diff
@@ -0,0 +1,136 @@
...
+    "file" : "orbitals"
+}
+
+]}
+EOF
+
+#----------BAGEL JOB-------------------------
+export BAGEL_NUM_THREADS=1
+export OMP_NUM_THREADS=1
+/home/srsen/bin/anaconda3/bin/mpirun -np 1 BAGEL $input.json > $input.out
```
danielhollas, line 112: Use the `SetEnvironment.sh` paradigm, take a look how it is done in the other interfaces. I would suggest to define the following two variables there:

```sh
export PATH=path/to/bagel:$PATH
BAGELEXE=BAGEL
MPIRUN=/home/srsen/bin/anaconda3/bin/mpirun
```

So then the line will be

```sh
source SetEnvironment.sh BAGEL
$MPIRUN -np $nproc $BAGELEXE $input.json &> $input.out
```
with `nproc` defined with all the other parameters.

interfaces/BAGEL-LZ/r.bagel
```diff
@@ -0,0 +1,136 @@
...
+
+if $( grep -q 'ERROR: EXCEPTION RAISED:  Max iteration reached during the second-order optimization.' $input.out ) ;then
```
danielhollas, line 114:
```suggestion
if `grep -q 'ERROR: EXCEPTION RAISED:  Max iteration reached during the second-order optimization.' $input.out` ;then 
```

interfaces/BAGEL-LZ/r.bagel
```diff
@@ -0,0 +1,136 @@
...
+    echo "ERROR: Algorithm could not converge!"
+    exit 2
+fi
+
+#####################################################################
+
+# NOW IT'S TIME TO COLLECT ALL THE DATA FOR ABIN
+
+# Extracting energy.
+cat ENERGY.out >> ../engrad.dat.$ibead
```
danielhollas, line 124:
```suggestion
cat ENERGY.out > ../engrad.dat.$ibead
```

interfaces/BAGEL-LZ/r.bagel
```diff
@@ -0,0 +1,136 @@
...
+#-about Density Fitting: http://www.molpro.net/info/2012.1/doc/manual/node133.html
+#-Easiest way to do DF-CASSCF is to use turbomole (Ahlrichs) Def2 basis sets (SVP,TZVP etc.)
+#-For Dunnging basis sets, DF bases are available, but you won't calculate NAC
+#-For Pople basis sets, you need to specify DF basis manually, probably some of the DF Ahlrichs basis sets.
+
+rm -f ../engrad.dat.$ibead ../nacm.dat
+
+#-How many atoms?
+natom=$(cat ../geom.dat.$ibead | wc -l )
+let natom1=natom+1
```
danielhollas, line 31: Delete these unused variables.

interfaces/BAGEL-SH/r.bagel
```diff
@@ -0,0 +1,142 @@
...
+timestep=$1
+ibead=$2
+input=input$ibead
+if [[ -z $3 ]];then
+   accu=7
+else
+   accu=$3
+fi
+
+# Definition of wavefunction
```
danielhollas, line 15: Apply suggestions from `BAGEL-LZ/r.bagel-lz`

interfaces/BAGEL-SH/r.bagel
```diff
@@ -0,0 +1,142 @@
+#!/bin/bash
```
danielhollas, line 1: Rename file to `r.bagel-sh`

interfaces/BAGEL-LZ/r.bagel
```diff
@@ -0,0 +1,136 @@
+#!/bin/bash
```
danielhollas, line 1: Rename file to `r.bagel-lz`

interfaces/BAGEL-SH/r.bagel.nacm
```diff
@@ -0,0 +1,148 @@
+#!/bin/bash
```
danielhollas, line 1: Rename to `r.bagel-sh.nacm`

interfaces/BAGEL-SH/r.bagel.nacm
```diff
@@ -0,0 +1,148 @@
...
+
+
+
+# END OF BAGEL INPUT
+
+# BAGEL JOB
+export BAGEL_NUM_THREADS=1
+export OMP_NUM_THREADS=1
+/home/srsen/bin/anaconda3/bin/mpirun -np 1 BAGEL $input.json > $input.out
+
```
danielhollas, line 116: Check the return value of this command, look how it is done in other interfaces.

```sh
if [[ $? -ne 0 ]];then
    echo "BAGEL exited with error. Check $input.out"
    exit 2
fi
```

danielhollas (changes requested):
Awesome, thanks so much to all who contributed to this interface! :+1:

interfaces/BAGEL-LZ/r.bagel
JanosJiri: Done

interfaces/BAGEL-LZ/r.bagel
danielhollas: I have confused myself here. `$(command)` and `` `command` `` are essentially equivalent. They are called [command substitution](https://www.gnu.org/software/bash/manual/html_node/Command-Substitution.html), and return the standard output (stdout) of the command. But `grep -q` will never print anything into its stdout anyway.

What we instead want to be testing is the _return value_ (or exit status) of the command. In linux, successful execution should always return 0, which (somewhat confusingly) is evaluated as true by the if statement. Anything other than 0 is false. One can check that by looking at the return values of the `false` and `true` commands.

```
$ /bin/true; echo $?
0
$ /bin/false; echo $?
1
```

You can also checkout the `EXIT STATUS` section of grep's manpage.

The original code still works, because the return value of the command substitution is the return value of the command itself. But it is superfluous, one can simply write

```
if grep -q 'ERROR: ...' $input.out; then

fi
```

Here's a little comic about various brackets in BASH that I found extremely useful.
https://wizardzines.com/comics/brackets-cheatsheet/

CC @JanosJiri

danielhollas:
@JanosJiri is this ready for a new round of review or not yet?

JanosJiri:
Yes, it's ready for review @danielhollas.

interfaces/BAGEL-LZ/r.bagel-lz
```diff
@@ -0,0 +1,210 @@
...
+# COMMENT: Therefore, I skipped changing covergence criteria during CASSCF routine. 
+
+
+####################################################################
+
+# NOW IT'S TIME TO COLLECT ALL THE DATA FOR ABIN
+
+# Extracting energy.
+cat ENERGY.out >> ../engrad.dat.$ibead
+cat ENERGY.out >> ../engrad_all.dat
```
danielhollas, line 199:
```suggestion
cat ENERGY.out > ../engrad.dat.$ibead
cat ENERGY.out > ../engrad_all.dat
```

interfaces/BAGEL-LZ/r.bagel-lz
```diff
@@ -0,0 +1,210 @@
...
+    "title" : "molecule",
+    "basis" : "$basis",
+    "df_basis" : "$df_basis",
+    "angstrom" : "true",
+    "geometry" : [
+EOF
+
+# section for geometry specification in json format
+for j in `seq 1 $natom`; do
+    atom=$(echo $(cat ../geom.dat.$ibead | head -n $j | tail -n 1) | awk '{print $1}')
```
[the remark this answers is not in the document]
danielhollas: Please simplify all of this with AWK. Something like this should work

```awk
awk -v natom="$natom" ' \
{printf('{ "atom" : %d, "xyz" : [ %.15f, %.15f, %.15f]}", natom, $1, $2, $3} \
NR != natom {print ","}
'
```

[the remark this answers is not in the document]
danielhollas: Please rewrite this with AWK, something like this should work

```awk
awk -v natom="$natom" '{printf "{ \"atom\" : %s, \"xyz\" : [ %.15f, %.15f, %.15f]}", $1, $2, $3, $4} NR != natom {print ","}END{print ""}'  ../geom.dat.$ibead >> $input.json
```

interfaces/BAGEL-LZ/r.bagel-lz
```diff
@@ -0,0 +1,210 @@
+#!/bin/bash
```
danielhollas, line 1: Please apply comments on this file to the files in BAGEL-SH as well.

interfaces/BAGEL-LZ/r.bagel-lz
```diff
@@ -0,0 +1,210 @@
...
+}
+
+]}
+EOF
+
+#----------BAGEL JOB-------------------------
+export BAGEL_NUM_THREADS=1
+export OMP_NUM_THREADS=1
+# specific launching of BAGEL - modify for different machines
+/home/srsen/bin/anaconda3/bin/mpirun -np 1 BAGEL $input.json > $input.out
```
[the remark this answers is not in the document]
danielhollas: Also it would be really great to get rid of this hard-coded path and use the SetEnvironment.sh approach. Note that here we also assume that BAGEL is in $PATH.

danielhollas (changes requested):
This is looking great now @JanosJiri! Some final small things, and then I'll test it locally with our BAGEL.

interfaces/BAGEL-LZ/r.bagel-lz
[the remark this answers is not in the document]
JanosJiri: Done

interfaces/BAGEL-LZ/r.bagel-lz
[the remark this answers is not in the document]
JanosJiri: Done, ready for review.

danielhollas:
Thanks @JanosJiri :clap: I'll do some testing locally (hopefully next week) and then we should be good to go!

danielhollas:
@JanosJiri I've cleaned up the interfaces in BAGEL-SH and tested XMS-CASPT2, MS-CASPT2 and CASSCF. I've been able to reduce a lot of duplication by moving stuff into functions in bagel.inp. If this looks good to you I'll apply the same thing to LZ and ground state BAGEL.

interfaces/BAGEL-SH/r.bagel-sh
```diff
@@ -0,0 +1,165 @@
...
+    if [[ -z ${target_state-} ]]; then
+      target_state=$ist
+    else
+      >&2 echo "ERROR: Invalid tocalc, cannot compute gradient for more than one state"
+      exit 2
+    fi
+  fi
+done
+
+cat >> $input.json << EOF
```
JanosJiri, line 54: You have a fucntion in bagel.inp to do this if I understood it correctly.

danielhollas:
Okay, I think I am done. :sweat_smile: I've cleaned up all the three interfaces and tested them on Hamilton. The scripts now also pass the `shellcheck` [BASH linter](https://www.shellcheck.net/), invoked as

```
shellcheck -a -x r.bagel*
```

@JanosJiri It would be great if you could do last tests on Prague clusters and then we're good to go. When testing, you can increase the `deltaE` parameter so that you also exercise the NACM calculation. Thanks!

danielhollas (approved):
I am going to merge this. We can always open a new PR if we run into any issues or have more improvements. 

Awesome work, very happy with what we have. :clap:

PR #159, opened 2023-04-12
Improvements of LZSH algorithm

JanosJiri:
This commit slightly modifies the LZ code to fix issues we encountered for several molecules.

1. LZ is computed just for the neighbouring states, not all that are available. Previously, the implementation allowed for unphysical hops like S1 --> S4.
2. Warning is printed out when the sum of hopping probabilities to the two neighbouring states is larger than one. ABIN somehow deals with that but LZ formula is no longer valid and would need a 3-state generalization. 
3. Warning is printed out when possible discontinuities are encountered. It is based on second derivatives. Testing on azobenzene and bilirubin showed it is able to warn the user when hops due to discontinuity happen.
4. Warning at the beginning of output if there are more than two states of the same multiplicity. In such cases, LZ works only when the states are well separated. This doesn't hold true for azobenzene, bilirubin, porphyrin and other molecules and LZ should not be used.
5. Check for dE>deltaE_lz is added at the beginning, otherwise it messes up with probabilities.

Btw, the simple velocity rescaling right be also a problem for LZSH but this was not modified.


src/landau_zener.F90
```diff
@@ -192,9 +199,24 @@ subroutine lz_hop(x, y, z, vx, vy, vz, fxc, fyc, fzc, amt, dt, eclas, chpot)
                call fatal_error(__FILE__, __LINE__, 'LZ probability > 1')
             end if
 
+            ! Adding check for false 3P-minima
+            ! only for significant probabilities, for tiny probabilities this does not make sense
+            ! e.g. for parallel states (the whole LZ does not make sense for this case)
+            if (prob(ist1) > 0.01) then
+                second_der_back = ((en_diff(2) - 2 * en_diff(3) + en_diff(4)) / dt**2)
+                der_check = abs((second_der - second_der_back) / second_der)
+                if (der_check > 0.3) then
```
danielhollas, line 208: Can you make a comment about how you came to this threshold? If I understand the code correctly, does this check if the second derivative changes more than 30%? Please add an explanation to make this super-clear.

src/landau_zener.F90
```diff
@@ -192,9 +199,24 @@ subroutine lz_hop(x, y, z, vx, vy, vz, fxc, fyc, fzc, amt, dt, eclas, chpot)
...
+                    write (*,*) "WARNING, possible discontinuity in PES! Check PES.dat!"
```
danielhollas, line 209: This seems like a rather serious issue that may lead to false hops. Should we rather stop instead of just issuing the warning?

src/landau_zener.F90
```diff
@@ -192,9 +199,24 @@ subroutine lz_hop(x, y, z, vx, vy, vz, fxc, fyc, fzc, amt, dt, eclas, chpot)
...
+                end if
+           end if
+
          end if
       end do
-      !write(stdout,*) "diff1",en_diff(1),"diff2",en_diff(2),"diff3",en_diff(3)
+
+      ! LZ warning
+      if (sum(prob) > 1) then
+          write (*,*) "* WARNING, sum of hopping probabilities > 1. Breakdown of LZ assumptions"
```
danielhollas, line 218:
```suggestion
          write (stdout, *) "WARNING: Sum of hopping probabilities > 1. Breakdown of LZ assumptions"
```

src/landau_zener.F90
```diff
@@ -553,7 +576,7 @@ subroutine lz_restout(fileunit)
 
       write (fileunit, *) istate_lz
       do ist = 1, nstate_lz
-         write (fileunit, '(3ES25.16E3)') en_array_lz(ist, 1), en_array_lz(ist, 2), en_array_lz(ist, 3)
+         write (fileunit, '(3ES25.16E3)') en_array_lz(ist, 1), en_array_lz(ist, 2), en_array_lz(ist, 3), en_array_lz(ist, 4)
```
danielhollas, line 579: This is a breaking change, right? Previous restart files will stop working after this?

src/landau_zener.F90
```diff
@@ -175,10 +175,17 @@ subroutine lz_hop(x, y, z, vx, vy, vz, fxc, fyc, fzc, amt, dt, eclas, chpot)
...
+         if (ist1 == ist) cycle 
+         ! only closest states are considered for hopping
+         if (ist1 > (ist + 1) .or. ist1 < (ist - 1)) cycle 
+
+         do ihist = 1, 4
             en_diff(ihist) = abs(en_array_lz(ist, ihist) - en_array_lz(ist1, ihist))
          end do
+
+         ! the energy condition is considered here, otherwise it messes up probabilities
+         if (abs(en_diff(2) * AUTOEV) > deltaE_lz) cycle
```
danielhollas, line 187: Could you make the comment at the top of the file where `deltaE_lz` is specified that it is in eV?

danielhollas (changes requested):
Cool! These look like very important fixes. I mostly have just a couple of clarifying questions.
In general, don't be afraid to spell things out a bit more in the code comments to make things super-clear for anybody who might look at this code. Thanks!

(note, the tests are not running because of the old Ubuntu version, I am fixing that in #160

src/landau_zener.F90
JanosJiri: Done

src/landau_zener.F90
JanosJiri: Done

src/abin.F90
```diff
@@ -68,6 +68,12 @@ program abin
    write (stdout, '(A)') 'Job started at: '//trim(get_formatted_date_and_time(time_start))
    write (stdout, *) ''
 
+   ! LZ warning for too many states
+   if (ipimd == 5 .and. (nsinglet_lz > 2 .or. ntriplet_lz > 2 )) then
```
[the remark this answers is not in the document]
danielhollas: Okay, fair enough, although I am not convinced that this particular warning is more important than other possible warnings. Perhaps in the future we should have some kind of warnings module that would print all warnings at the end of init.

src/landau_zener.F90
danielhollas: Okay, cool. It would be nice to add error handling for reading and print a meaningful error, but not neccessary for this PR. I will also have a breaking change to fix restart for FSSH. Perhaps we need some kind of versioning for the restart file to make these kind of changes more explicit.

src/landau_zener.F90
danielhollas: Thanks! Though it seems you haven't pushed your changes. btw: please take a look at #160, I need to merge that one first to make the tests work again on Github. Once it is merged, you can either rebase your branch or merge the master into it (`git merge master` or `git rebase master`). In the latter case you would need to force push.

src/landau_zener.F90
JanosJiri: Done

src/landau_zener.F90
JanosJiri: This is a bit problematic since it can be a steep conical intersection. I added a section with another condition, where for change bigger than 130% we set the probability of hopping to 0 artificially because we have an unphysical change of curvature and we have almost certainly a discontinuity.

src/landau_zener.F90
JanosJiri: I merged #160 into master, then merged with this branch and pushed here. All the modifications of LZSH are in the 3. commit,

danielhollas: @JanosJiri thanks! It looks like the last commit introduced a build failure, can you take a look?

danielhollas (changes requested):

src/landau_zener.F90
danielhollas: Okay, I would personally be more conservative and end stop the simulation for such a large discontinuity, but leaving that up to you.

src/landau_zener.F90
JanosJiri: Typo in one variable, now it is fixed and it's working. At least on Krypton.

danielhollas:
@JanosJiri are you compiling locally with MPI? Looks like the TERAPI-LZ tests are failing.

JanosJiri:
@danielhollas No, I didn't compile it with MPI. I had general issues with MPI here so I even did not try.

danielhollas:
@JanosJiri anything I can help with to get this over the finish line?

JanosJiri:
> @JanosJiri anything I can help with to get this over the finish line?

@danielhollas I know the problem is in the TERAPI-LZ test but I was not able to compile ABIN with MPI here. So I don't see where exactly it failed. I just know it did not hop in the test suite. I will try to compile it again but I won't have time to do it soon.

danielhollas:
Cool, thanks! In terms of compiling with MPI, I'd recommend using the `dev_scripts/install_mpich.sh` to compile your own MPICH version locally, and then rerunning configure with `./configure --mpi /path/to/mpi`. This is what we use on Github CI so it should work.

danielhollas:
@JanosJiri I've looked at the failing test `TERAPI-LZ`, and the issue is that in the original test the system hops from 3rd to 2nd state in the third timestep, but with the new code the output in `abin.out` says 

```
 Three-point minimum (03->02) dE/a.u.   4.2773499999952946E-003 Probability:  0.95692859086662962
 ERROR: Change of curvature --> discontinuity in PES!
 Probability set to 0!
```

I am not sure if this is expected, since the PES seems fine to me at first glance. Here are the curves between S2 and S1 states
image

Zoomed out view that includes S0

image

(of course this is a synthetic test but the curves don't seem unrealistic to me)

JanosJiri:
> @JanosJiri I've looked at the failing test `TERAPI-LZ`, and the issue is that in the original test the system hops from 3rd to 2nd state in the third timestep, but with the new code the output in `abin.out` says
> 
> ```
>  Three-point minimum (03->02) dE/a.u.   4.2773499999952946E-003 Probability:  0.95692859086662962
>  ERROR: Change of curvature --> discontinuity in PES!
>  Probability set to 0!
> ```
> 
> I am not sure if this is expected, since the PES seems fine to me at first glance. Here are the curves between S2 and S1 states image
> 
> Zoomed out view that includes S0
> 
> image
> 
> (of course this is a synthetic test but the curves don't seem unrealistic to me)

@danielhollas Thanks for looking at it and also sorry for taking such a long time with it. I believe the problem is in the test, not in the code. The 'smoothness' check I implemented compares the current second derivative of the energy difference with the previous one. However, since this hops on the third step of the algorithm, there is no previous second derivative to compare with. Or rather the 4th energy in the array is still zero generating an artificial second derivative. So I believe this can be solved by adding one more data point before the hopping step. All should be fine then. I will have a look but in a three weeks when I will be back from holidays.

danielhollas:
@JanosJiri Thanks for fixing the test!

>  I believe the problem is in the test, not in the code. The 'smoothness' check I implemented compares the current second derivative of the energy difference with the previous one. However, since this hops on the third step of the algorithm, there is no previous second derivative to compare with. Or rather the 4th energy in the array is still zero generating an artificial second derivative.

I am not sure I agree that the problem was in the test though (or at least I wouldn't phrase it that way). Before this PR it was perfectly legal to jump on the third step (but not in the first two steps). After this change, we're essentially disallowing any hopping before 4th steps. I guess that is okay, but I am not very happy that there is a misleading warning emitted if the system wants to jump in the third step (i.e. user might think there was discontinuity, even though in reality there wasn't). 

@JanosJiri Let me know your thoughts. I am happy to merge this as is, and we can improve upon this later. :rocket:

JanosJiri:
> @JanosJiri Thanks for fixing the test!
> 
> > I believe the problem is in the test, not in the code. The 'smoothness' check I implemented compares the current second derivative of the energy difference with the previous one. However, since this hops on the third step of the algorithm, there is no previous second derivative to compare with. Or rather the 4th energy in the array is still zero generating an artificial second derivative.
> 
> I am not sure I agree that the problem was in the test though (or at least I wouldn't phrase it that way). Before this PR it was perfectly legal to jump on the third step (but not in the first two steps). After this change, we're essentially disallowing any hopping before 4th steps. I guess that is okay, but I am not very happy that there is a misleading warning emitted if the system wants to jump in the third step (i.e. user might think there was discontinuity, even though in reality there wasn't).
> 
> @JanosJiri Let me know your thoughts. I am happy to merge this as is, and we can improve upon this later. 🚀

First of all, I agree it is not ideal but there shouldn't be a big problem. The problem is that it won't allow hopping the 2nd step of simulation (not 3rd, 3rd is propagation after hopping) because it has no memory of second derivatives. This should not be a problem when launching the simulations, because in LZ, you should not hop so early, the theory isn't derived for that (it assumes long propagation in the coupling region). The only problem might occur when restarting simulation.

I would merge it and later we can add the second derivative memory to the restart files, which is the only problem I see.

danielhollas (approved):
Okay, sounds good, thanks @JanosJiri

danielhollas:
Ugh, I thought I had already merged this. Merging now, apologies for the delay.

PR #160, opened 2023-04-12
GHA: Update Ubuntu version to 20.04

danielhollas:
Github actions removed the runner for Ubuntu 18.04, which is why ABIN CI tests were not running. Here I update all jobs to Ubuntu 20.04. 

~~This unfortunately means we can no longer test with gfortran-7.
@JanosJiri for your PRs, please always manually test on Argon or Neon with GCC 7.3.~~

EDIT: I figured out how to install GCC-7 so we're good here.

### DIFF
--- a/.github/workflows/gfortran.yml
+++ b/.github/workflows/gfortran.yml
@@ -177,7 +177,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-         gcc_v: [11, 12]
+         gcc_v: [9, 11, 12]
     env:
       FC: gfortran
       GCC_V: ${{ matrix.gcc_v}}
--- a/.github/workflows/gfortran.yml
+++ b/.github/workflows/gfortran.yml
@@ -11,7 +11,7 @@ on:
 
 env:
   # FFLAGS for building ABIN, applicable for most jobs
-  ABIN_FFLAGS: -O0 -fopenmp --coverage -fprofile-abs-path -fno-inline -fcheck=all -ffpe-trap=invalid,zero,overflow,denormal -fimplicit-none -Wall -Wno-integer-division -Wno-maybe-uninitialized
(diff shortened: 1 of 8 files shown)


danielhollas:
Code coverage is not working for certain builds (FFTW, Plumed, OpenMPI), working on it...

danielhollas:
There are some residual coverage changes that I haven't been able to resolve, which mainly come from the FFTW build (so affects normal mode PIMD/PIGLET). It looks like gcc/gcov versions >7 have some bugs which means some statements (e.g. `allocate`/`deallocate`) look like not covered even though they clearly are. For other builds it's not a problem since we still build with GCC7 at least once. But FFTW build must run on GCC version that is default for the given Ubuntu version, because we install `libfftw` via apt. If we wanted to switch the FFTW build to GCC-7, we would need to compile `libfftw` ourselves, otherwise there's a conflict. This seems like too much work so we will take the coverage hit for now, and hope it will get resolved in future GCC versions. I have also tested various compiler options but none of them fixed the situation.

Note that some coverage changes seem to be real, specifically the lines that became "partially" covered, see the code on codecov for details. These come from the if statements that are not fully exercised in terms of "branch coverage", meaning that the if statement is either always true, or always false, end hence not all possible code branches are exercised.

danielhollas:
I have enabled extra bound checks and runtime memory leak detection on some of the builds. (I couldn't enable it for all builds because it messes up with the code coverage). This uncovered one minor out-of-bounds array access in NHC initialition, which was however not critical since it was happening only in the error path. I also uncovered a memory leak in the pFUnit library that we use for unit tests, which I reported here: https://github.com/Goddard-Fortran-Ecosystem/pFUnit/issues/419

JanosJiri (approved):
I don't understand the gfortran.yml and codecov.yml staff but the rest of it seemed OK for me.

danielhollas:
@JanosJiri thanks for taking a look. I'll walk you through the CI settings once you're back in Bristol.

In terms of merging, for PRs like these with a large number of commits (56!) it is better to use the "Squash and merge" strategy, which means that all the commits will be squashed to one, resulting in much cleaner commit history. I will fix this up on master (but will have to force push to master, which is not ideal, you might need to fix up your local repo if you've pulled master already).

PR #161, opened 2023-04-18
Add gcov args that eT uses

danielhollas:


danielhollas:
This did not seem to help

PR #164, opened 2023-06-19
Use usleep to wait for TeraChem

### LINKED ISSUE #163: Switch to POSIX sleep when waiting for MPI communication from TeraChem
The current strategy of just shelling out to `sleep` unix command is not great.

https://cyber.dabamos.de/programming/modernfortran/sleep.html
https://github.com/scivision/fortran-sleep/tree/main

danielhollas:
When communicating with TeraChem via MPI, we're sleeping for 50 miliseconds. Otherwise, ABIN busy-waits and consumes the whole CPU.

Here is a better implementation of this behaviour, using the system call function `usleep`, instead of calling to shell. 
Closes #163

### DIFF
--- a/src/fortran_interfaces.F90
+++ b/src/fortran_interfaces.F90
@@ -55,6 +55,7 @@ module mod_interfaces
       !   integer(C_INT), intent(in) :: natom, walkmax, watpot
       !end subroutine
 
+      ! https://cyber.dabamos.de/programming/modernfortran/sleep.html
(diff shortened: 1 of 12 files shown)


JanosJiri (approved):
Seems fine to me.

danielhollas:
@yoricklassmann could you please take a look as well? Thanks!

yoricklassmann:
Looks good to me.

PR #165, opened 2023-07-05
Bring back TeraChem file interface

danielhollas:
The Amber MPI interface in TC was removed in 2021. We've used this interface for ground state MD.
Instead, we now have the TCPB interface, but still in an experimental stage. Therefore, for now we bring back the simple file interface as a quick hassle free way to run MD with TeraChem.
WARNING: For small / medium sized molecules, you will lose a lot of time during repeated GPU initialization!

### DIFF
--- a/interfaces/TERACHEM/r.terachem
+++ b/interfaces/TERACHEM/r.terachem
@@ -1,16 +1,19 @@
 #!/bin/bash
-cd $(dirname $0)
-timestep=$1
+set -u
+
+# This script is called by ABIN as:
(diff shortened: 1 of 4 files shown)


danielhollas:
@yoricklassmann could you please review this? Thanks! (no rush)

danielhollas:
@JanosJiri would you mind giving this a test on Prague clusters? Thanks!

danielhollas:
I tested again on Jackdaw with short NVE dynamics, should be good.

PR #166, opened 2023-10-09
Add Schwenke's gas phase H2O analytical potential

danielhollas:


src/force_h2o.F90
```diff
@@ -0,0 +1,114 @@
...
+      end do
+      Eclas = Eclas / nbeads
+
+      call numerical_forces(x, y, z, fx, fy, fz, Epot, natom, nbeads)
+
+   end subroutine force_h2o_schwenke
+
+   ! TODO: Implement numerical forces generally for all potentials
+   ! For now, they can be implemented here and hardcoded for a specific H2O potential
+   subroutine numerical_forces(x, y, z, Epot, fx, fy, fz, natom, nbeads)
```
danielhollas, line 99: @mbarnfield63 Here's where you can start with your implementation.
Ultimately we'll want to have the implementation of numerical forces generalized and in a separate module, but that's not important at this stage.

danielhollas (commented):
@mbarnfield63 It would be great if you could go through the changes in this PR, and make a comment where things are not clear to you. Focus on file `force_h2o.F90`. The actual potential is in `h2o_schwenke.f`, but that has been taken essentially verbatim from the paper (and we already went through it together once). 
Changes to the other files are not that important, they are mostly just hooking up the new potential trough ABIN. (and there are a couple of unrelated changes as well).

I am going to merge this to the master branch, which you can then pull-in and start your own branch on your fork for the numerical forces.
 
Note that I have renamed the module and function names a bit to make them consistent, so your scan script would need to be updated accrodingly.

PR #167, opened 2023-10-30
Vendor fprettify autoformater and add to CI

danielhollas:


JanosJiri (approved):
Seems to do nice formatting of the code.

PR #174, opened 2024-02-26
Schwenke numerical

mbarnfield63:
Implemented numerical_forces subroutine using forward differences for numerical force calculation.


.gitignore
danielhollas, line 1: Please revert changes to this file

sample_inputs/input.in.cmd
danielhollas, line 1: ditto

src/cvrqd_scan
danielhollas, line 1: Please remove binary file

src/force_h2o.F90
```diff
@@ -99,19 +100,81 @@ end subroutine force_h2o_schwenke
...
-      real(DP), intent(in) :: y(natom, nbeads)
-      real(DP), intent(in) :: z(natom, nbeads)
-      real(DP), intent(inout) :: fx(natom, nbeads)
-      real(DP), intent(inout) :: fy(natom, nbeads)
-      real(DP), intent(inout) :: fz(natom, nbeads)
-      ! This is the energy for the currrent geometry that has already been calculated
-      real(DP), intent(in) :: Epot(nbeads)
+
+   subroutine numerical_forces(x, y, z, fx, fy, fz, Epot, natom, nbeads)
+      real(DP), intent(in) :: x(3, 1)
```
danielhollas, line 105:
```suggestion
      real(DP), intent(in) :: x(natom, nbeads)
```
same below

src/force_h2o.F90
```diff
@@ -99,19 +100,81 @@ end subroutine force_h2o_schwenke
...
-      call fatal_error(__FILE__, __LINE__, 'Numerical forces not yet implemented!')
-   end subroutine numerical_forces
+      ! Create new copies of arrays
+      real(DP) :: x_new_forward(3, 1)
+      real(DP) :: y_new_forward(3, 1)
+      real(DP) :: z_new_forward(3, 1)
 
-end module mod_force_h2o
+      ! This is the energy for the currrent geometry that has already been calculated
+      real(DP), intent(in) :: Epot(nbeads) !nbeads
```
danielhollas, line 119:
```suggestion
      real(DP), intent(in) :: Epot(nbeads)
```

src/force_h2o.F90
```diff
@@ -99,19 +100,81 @@ end subroutine force_h2o_schwenke
...
+                  case (2)
+                     y_new_forward(i,1) = y_new_forward(i,1) + delta
+                  case (3)
+                     z_new_forward(i,1) = z_new_forward(i,1) + delta
+               end select
+
+               ! Calculate the energy for the forward perturbed geometry
+               call get_internal_coords(x_new_forward, y_new_forward, z_new_forward, j, new_rOH1, new_rOH2, new_aHOH_rad)
+               
+               new_rij(j, :) = [new_rOH1, new_rOH2, new_aHOH_rad]
```
danielhollas, line 160: Huh, I've never seen this syntax, suprised that it works. Can you provide a link where you find this syntax explained?

src/force_h2o.F90
```diff
@@ -99,19 +100,81 @@ end subroutine force_h2o_schwenke
...
+               
+               call h2o_pot_schwenke(new_rij, Epot_delta, nbeads)
```
danielhollas, line 162: I don't think this call is quite correct. You do not want to calculate the energy for multiple beads, but instead only for the bead that you displaced. So I think `Epot_delta` should be an array of size one. You could do it in a different way, but this one seems the easiest.

```suggestion
               call h2o_pot_schwenke(new_rij, Epot_delta, nbeads)
```

danielhollas: It might be that in the end you are looking at the correct energies below. But at the very least this is very inefficient, since you end up calculating energies for multiple beads and then only use one of them below.

src/force_h2o.F90
```diff
@@ -99,19 +100,81 @@ end subroutine force_h2o_schwenke
...
+
+               Eclas_plus = Epot_delta(j)
+
+               ! Calculate the numerical force
+               select case (k)
+                  case (1)
+                     fx(i,1) = -(Eclas_plus - Eclas_orig) / delta
```
danielhollas, line 169: I think the indexing here is wrong.

danielhollas (changes requested):
Just a quick review. It seems there are some indexing errors for the PIMD implementation of forces.

src/force_h2o.F90
```diff
@@ -69,7 +70,7 @@ subroutine force_h2o_schwenke(x, y, z, fx, fy, fz, Eclas, natom, nbeads)
       integer, intent(in) :: natom, nbeads
       ! Internal water coordinates
       real(DP) :: rOH1, rOH2, aHOH_rad
-      real(DP) :: rij(nbeads, 3)
```
danielhollas, line 72: I don't think this change is correct. Note that you're iterating over the first index on lines 78-83

danielhollas:
@mbarnfield63 could you please push to GitHub the code that you're actually using to run the production MD?

src/force_h2o.F90
danielhollas: @mbarnfield63 I believe this still needs fixing as well.

danielhollas (approved):

PR #177, opened 2024-03-13
Cleanup create_trajectories.sh with shellcheck

danielhollas:
Mostly cosmetic changes and cleanup.

One functional change: It is now possible to specify `submit_command=bash` if the trajectories are to be run directly on a machine (e.g. Squirrel or Hamilton nodes) instead of submitting to queing system. In this case, the launch script `r.abin` doesn't need to be provided in the template directory.

### DIFF
--- a/utils/create_trajectories.sh
+++ b/utils/create_trajectories.sh
@@ -200,12 +200,22 @@ done
 # Submit jobs
 k=1
 if [[ -n "${submit_command-}" ]];then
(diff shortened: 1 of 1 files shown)


JanosJiri (approved):
Looks all good to me

danielhollas:
Cheers @JanosJiri.

I'll merge after Jack tests it out on Jackdaw.

PR #178, opened 2024-05-08
NaI analytic potential for FSSH

JanosJiri:
Analytic potential for NaI implemented for surface hopping dynamics. The original potential is adiabatic but was adiabatized in the code. Everything even with the source articles is described in the code. 3ps of dynamics can be done in 8 seconds when one is not writing too much on the disk. Accessing the disk is the most demanding part of the simulation. The implementation was compared to the exact QD.
plot


src/potentials_sh.F90
```diff
@@ -0,0 +1,168 @@
...
+   public
+   private :: nai
+
+   ! We use derived types to encapsulate potential parameters
+   ! https://fortran-lang.org/learn/quickstart/derived_types
+
+   ! Parameters for the NaI model as defined in the following articles
+   ! https://doi.org/10.1063/1.4919780
+   ! https://doi.org/10.1063/1.456377
+   ! These articles contain diabatic basis while here we use adiabatic
```
danielhollas, line 19: Could you explain here a bit more how you got from the diabatic to adiabatic parameters?

src/potentials_sh.F90
```diff
@@ -0,0 +1,168 @@
+! Various analytical potentials for surface hopping:
+! - NaI potential and couplings in adiabatic basis
+
+! Created by Jiri Janos
+
+! User-definer parameters are set input section 'system'
```
danielhollas, line 6: I think the parameters are currently hard-coded right?
```suggestion
```

src/potentials_sh.F90
danielhollas, line 1: Let's rename this file to something more specific, like `src/nai_potential.F90`. If we implemented more potentials like this in the future they should go to their own modules.

src/surfacehop.F90
```diff
@@ -287,6 +287,7 @@ subroutine get_nacm(pot)
 
       ! In TeraChem SH interface, we already got NACME
       if (pot == '_tera_') return
+      if (pot == '_nai_') return ! for NaI model, the couplings are calcualted with forces
```
danielhollas, line 290:
```suggestion
      if (pot == '_nai_') return ! for NaI model, the couplings are calculated together with forces
```

src/surfacehop.F90
```diff
@@ -611,16 +612,14 @@ subroutine surfacehop(x, y, z, vx, vy, vz, vx_old, vy_old, vz_old, dt, eclas)
 
       ! First, calculate NACME
       if (inac == 0) then
-
          ! For TeraChem MPI / FMS interface, NAC are already computed!
-         if (pot /= '_tera_') then
+         if (pot /= '_tera_' .and. pot /= '_nai_') then
```
danielhollas, line 616: Yeah, this is not great, sorry that you had to deal with this. I'll try to refactor this at some point.

danielhollas (changes requested):
This looks great, amazing work @JanosJiri! :clap: Really excited for this. 

I'll have a closer look at the implementation later, but it looks good overall from high level. 
Thank you for a very clean code and comments. 

Besides some minor stuff I noticed, it would be great if you could add at least one basic test.

Also, please run `./autoformat.py` which should autoformat the code. You might need to split some of the long lines manually, since the formater isn't clever enough for that.

JanosJiri:
Fprettify applied and one test added. I also processed all the changes requested except for one where I have a question for you. Once we resolve this, the code will be ready for merging.

tests/test
danielhollas, line 1: I guess this file was added by mistake?

danielhollas:
Excellent, thanks!

> Fprettify applied and one test added. I also processed all the changes requested except for one where I have a question for you. 

Hmm, I don't see the question anywhere. Did you post it on Github? (Github review UI is sometimes super confusing so perhaps I just missed it)

tests/test
JanosJiri: Yes.

danielhollas:
Thanks, the comment is marked as "pending", you'll need to "Finish the review" to publish it. 
I'll answer in the afternoon.

src/potentials_sh.F90
danielhollas: Sorry for a delayed replied. I am on a fence on this --- `potentials.F90` grew over time and contains much simpler potentials. But let's keep the name for now, we can always change it later. Thanks!

danielhollas:
@JanosJiri apologies for the delay while I was away. This looks really great, thanks for adding a test and extensive comments. Just one question, could you reduce the number of steps in the test? We generally try to keep the number of steps to the absolute minimum for numerical stability, to reduce the chance that the test will break with different compiler versions for example. Also, the test now takes more than 5s, which might not seem like much, but since we have a lot of them it adds up. 

When you modify the `input.in` in the test, you can quickly regenerate the reference files with 

```console
make makeref TEST=SH_NaI
```

In most tests with only do one step. Maybe here we want a bit more so we can test the value of NACM? But I think you anyway see it in `dotprod.dat`, right?

danielhollas (approved):
:rocket:

PR #182, opened 2024-10-24
Disable spline potential

danielhollas:
ABIN has an implementation of a numerical 1D potential on top of user-defined grid of points (`pot="_splined_grid_"` ). We use cubic splines to interpolate between the points.

Unfortunately, the cubic spline routines (originally taken from QDYN) are from Numerical recipes and we don't have license for them. Until we implement our own cubic spline, this commit remove the NR routines and disables the `_spline_` potential and associated tests.


src/force_spline.F90
```diff
@@ -1,8 +1,10 @@
 ! 1D user-defined numerical potential with cubic splines
 ! Original code taken from QDYN
 ! https://github.com/PHOTOX/qdyn
```
JanosJiri, line 3: Just a comment, the cubic splines are not implemented in QDyn anymore. It's probably still in some of the old commits.

JanosJiri (approved):
Looks good, I was just surprised you can't use splines from numerical recipes.

src/force_spline.F90
danielhollas: Good to know. What do you use instead?

JanosJiri: I use Python scripts to prepare the potential.

PR #185, opened 2024-11-06
Add 'ignore_state' in Launda-Zener hopping

danielhollas:
Another small bug (or perhaps rather a missing feature) found by Jack Taylor. Sometimes we want to calculate more electronic state (e.g. to make CASSCF more stable) but do not want to consider them in hoping algorithms. A prototypical example is simulating ethylene with SA3-CAS(2,2) starting from S1 and wanting to prevent hops to S2.

Also a minor refactoring to make more LZ stuff private to the `mod_lz` module.

### DIFF
--- a/src/abin.F90
+++ b/src/abin.F90
@@ -25,7 +25,7 @@ program abin
       & nwrite, nstep, it, inormalmodes, istage, irest, STOP_SIMULATION
    use mod_init, only: init
    use mod_sh, only: surfacehop, sh_init, get_nacm, move_vars
-   use mod_lz, only: lz_hop, en_array_lz, lz_rewind, nsinglet_lz, ntriplet_lz
+   use mod_lz, only: lz_hop, en_array_lz, lz_rewind
    use mod_kinetic, only: temperature
    use mod_utils, only: del_file, archive_file
    use mod_transform, only: initialize_pi_transforms, &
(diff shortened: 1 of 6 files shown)


src/abin.F90
```diff
@@ -68,12 +68,6 @@ program abin
    write (stdout, '(A)') 'Job started at: '//trim(get_formatted_date_and_time(time_start))
    write (stdout, *) ''
 
-   ! LZ warning for too many states
-   if (ipimd == 5 .and. (nsinglet_lz > 2 .or. ntriplet_lz > 2)) then
```
danielhollas, line 72: Moved this check to `lz_init`

src/landau_zener.F90
```diff
@@ -177,7 +241,8 @@ subroutine lz_hop(x, y, z, vx, vy, vz, fxc, fyc, fzc, amt, dt, eclas, chpot)
 
       do ist1 = ibeg, iend
          if (ist1 == ist) cycle
-         ! only closest states are considered for hopping
+         if (ist1 == ignore_state) cycle
```
danielhollas, line 244: This is the important bit!

tests/test.sh
```diff
@@ -227,27 +227,31 @@ do
...
       else
-         $ABINEXE > $ABINOUT 2>&1 || true
+         $ABINEXE > $ABINOUT 2>&1
       fi
 
-      #for testing restart
-      if [[ -e input.in2 ]];then
-         $ABINEXE -i input.in2 >> $ABINOUT 2>&1 || true
+      # for testing restart, only execute if previous run did not fail
+      if [[ -f input.in2 && $? -eq 0 ]]; then
```
danielhollas, line 253: It was annoying that even if a test failed, we were still running this second simulation (which is bound to fail as well)

JanosJiri (approved):
It looks good, I don't see any problems.

suchanj (approved):
That's a clever feature!

PR #186, opened 2024-11-07
Baeck-An couplings for SH

JanosJiri:
Approximate Baeck-An couplings are calculated based on the history of PESs and allow to run FSSH without explicitly calculated couplings.


JanosJiri:
@danielhollas could you please help me with the **GFortran CI / Basic build (7) (pull_request)** test failing all the time? I have all the tests passed locally and the other versions pass here too. I'm not sure what is the exact problem there. Otherwise it should be working Baeck-An implementation which I'm currently testing in Prague.

danielhollas:
This is the important bit

> line 210:  5054 Floating point exception(core dumped)

This likely means a bug in the code which is revealed by adding `--check=all` to FFLAGS (which is done only on GCC-7 build for *reasons*)

https://github.com/PHOTOX/ABIN/blob/094ba89da07cbbe983d2f4562eb182df203ffdbc/.github/workflows/gfortran.yml#L71

Try adding `-g check=all` to your local `make.vars` and hopefully you'll be able to reproduce the issue

src/files.F90
```diff
@@ -173,7 +173,7 @@ subroutine files_init(isbc, phase, ndist, nang, ndih)
          open (UPOP, file=chfiles(UPOP), access=chaccess, action='write')
          open (UPROB, file=chfiles(UPROB), access=chaccess, action='write')
          open (UPES, file=chfiles(UPES), access=chaccess, action='write')
-         open (UNACME, file=chfiles(UNACME), access=chaccess, action='write')
+         !note, UNACME is now opened in mod_sh since we don't use it for inac==1
          open (UDOTPROD, file=chfiles(UDOTPROD), access=chaccess, action='write')
```
danielhollas, line 177: Let's put all of this (including debug files below) into a separate function `sh_files_init()` or something like that?

src/surfacehop.F90
```diff
@@ -92,8 +96,8 @@ module mod_sh
    ! for ethylene 2-state SA3 dynamics
    integer :: ignore_state = 0
 
-   namelist /sh/ istate_init, nstate, substep, deltae, integ, inac, nohop, phase, decoh_alpha, popthr, ignore_state, &
-      nac_accu1, nac_accu2, popsumthr, energydifthr, energydriftthr, adjmom, revmom, &
+   namelist /sh/ istate_init, nstate, substep, deltae, integ, couplings, nohop, phase, decoh_alpha, popthr, ignore_state, &
```
danielhollas, line 99: Thanks for introducing the new keywords! 

I think we should keep the old variables here for backwards compatibility (so that we don't break existing input files).

Can you also update `sample_inputs/input.in.sh`

JanosJiri:
> This is the important bit
> 
> > line 210:  5054 Floating point exception(core dumped)
> 
> This likely means a bug in the code which is revealed by adding `--check=all` to FFLAGS (which is done only on GCC-7 build for _reasons_)
> 
> https://github.com/PHOTOX/ABIN/blob/094ba89da07cbbe983d2f4562eb182df203ffdbc/.github/workflows/gfortran.yml#L71
> 
> Try adding `-g check=all` to your local `make.vars` and hopefully you'll be able to reproduce the issue
> 
> > Otherwise it should be working Baeck-An implementation which I'm currently testing in Prague.
> 
> 🎉

Thanks for the advice. When I added `--check=all` to my FFLAGS like
```
FFLAGS = -O2 -fopenmp -fimplicit-none -Wall -Wno-integer-division -Wno-maybe-uninitialized --check=all
```
it again compiled without any problems. The flag `-g check=all` doesn't work when I add it to my make.vars. What confuses me is that it fails for only one version of GFortran and GCC, the others are ok. Plus, the error appeared when I modified the initial velocity file in the test. It was ok before.

danielhollas:
Did you run the test with the new build? These are runtime checks, not compile-time checks. You also want to disable optimizations with -O0, this should.

Only gcc7 on CI has these flags, have a look at the yaml file definition.

JanosJiri:
> Did you run the test with the new build? These are runtime checks, not compile-time checks. You also want to disable optimizations with -O0.

Yes, I ran the tests and everything passed. I tried again also with -O0, everything passed fine. I just can't get the error locally. I will at least try to go through the last commit if there is anything problematic.

src/surfacehop.F90
JanosJiri: > Can you also update sample_inputs/input.in.sh

Done!

About keeping `inac` and `adjmom`, if I keep them in the namelist, someone can set them but they will still be reassigned by `couplings` and `mom_adjust`, which have their default values. I'm afraid this would cause more problems. I've never seen input where these two were used so I wouldn't be worried about backwards compatibility of things done in Prague. When you changed the name for decoh_alpha, it was not backwards compatible and still people quickly adapted (again at least here in Prague). So I would stick just to the new keywords, yet I leave the executive decision up to you. :)

JanosJiri:
> What Gfortran version are you using?

gcc version 10.2.1 20210110 (Debian 10.2.1-6) 

> ```FFLAGS = -O0 -g -fopenmp -ffpe-trap=invalid,zero,overflow,denormal ```

This worked, thanks a lot! Now it should be all working.

src/surfacehop.F90
danielhollas: What you could do is to set the `couplings` keyword to some default value, and if the user doesn't change it in the input, don't overwrite the old keywords. I think I already do that for `ipimd` and `mdtype`. But I don't feel too strongly.

src/files.F90
JanosJiri: Done in the latest commit.

danielhollas:
Did you run the test with the new build? These are runtime checks, not compile-time checks. You also want to disable optimizations with -O0, this should.

danielhollas:
Awesome work! I'll have a look sometimes this week (I am currently in Switzerland on AiiDA coding week so not sure when I get to it).

It would be interesting to see how Landau works on the NAI model.

sample_inputs/input.in.sh
```diff
@@ -22,6 +22,8 @@ inose=0,		! NVE ensemble (thermostat turned OFF)
 &sh
 istate_init=2,		! initial electronic state (1 is ground state)
 nstate=3,		! number of electronic states
+couplings='analytic',	! non-adiabatic coupling terms 'analytic', 'baeck-an', 'none'
+mom_adjust='nac'        ! momentum adjustment along either 'nac' or 'velocity'
```
suchanj, line 26: I presume this is a velocity rescaling either along NAC or by scaling the whole vector? Mom immediately associates maximum overlap for me so it might be better to pick more suitable name but I don't insist on it. (vel_rescale=nac/all?)

suchanj (approved):

sample_inputs/input.in.sh
JanosJiri: Yes, it's the velocity rescaling. I see the point, we can rename it to `vel_rescale`, I will just wait for Dan's opinion so that I'm not changing it twice. Thanks!

danielhollas: I agree that `vel_rescale` is better, we could even do `velocity_rescaling` while we're at it, it's not like we're paying for the extra characters. :-) And it's not something that users will type often (and internally in ABIN we can use a shorter name. It's good anyway to decouple the names in the input file from the names in the code, the current design is not ideally. 

Perhaps a more interesting question is how to name the values.

- for the isotropic velocity rescaling, using all available kinetic energy, I would perhaps use `isotropic`. I am not aware of any widely used term for this, but perhaps it would be good to see how Mario calls it (for example in [Recommendations for Velocity Adjustment in Surface Hopping][1]).
- for the rescaling along NAC, the current implementation is somewhat subtle -- we first try to rescale along NAC, and if that fails, we rescale isotropically. So I think the option should be named as `nac_then_isotropic` or something like that, to make it obvious. 

The `nac_then_isotropic` approach made sense at the time (ten years ago) to reduce the number frustrated hops. It was also what Newton-X did at the time by default (I am not sure what they do now). However, as [Mario recently shown][1] quite conclusively, this can lead to artifacts for larger molecules, which will always have enough total kinetic energy to hop back. So as a follow-up to this PR, it would be great to implement the true `nac` option, and perhaps consider it as a safer default.

[1]: https://pubs.acs.org/doi/abs/10.1021/acs.jctc.3c01159

JanosJiri: Thanks for the comment, `velocity_rescaling` sounds good, I will use it. Regarding the options, I'm not a big fan of `isotropic` since I always saw in papers "rescaling along the momentum/velocity vector", which is also the case in Mario's paper you mentioned. Thus, I would go with `velocity` and `nac_then_velocity`, but I don't have a strong opinion. 

Regarding the rescaling with NAC and then possibly velocity, this is interesting and I believe we should add also pure NAC rescaling or at least document well the current version. I'm certain most of the papers from Prague claim rescaling along NAC not knowing there is velocity rescaling in case NAC fails. 

Following this topic, it might be worth making a new release of ABIN since we are all citing the old version. There has been quite a bit of development since then.

danielhollas: > Following this topic, it might be worth making a new release of ABIN since we are all citing the old version. 

yep, it's been on my mind for some time, but was never a priority. Should we target end of the year for new release? :-) I'll create a separate issue for discussion.

> There has been quite a bit of development since then.

And it's awesome to see that most of the new feature come from other people and not me. :relaxed: 

> I'm certain most of the papers from Prague claim rescaling along NAC not knowing there is velocity rescaling in case NAC fails.

Yeah, that's on me for not documenting things properly. I am pretty sure that in at least one paper I did explicitly wrote it down, but that's not an excuse.

> I'm not a big fan of isotropic since I always saw in papers "rescaling along the momentum/velocity vector", which is also the case in Mario's paper you mentioned. Thus, I would go with velocity and nac_then_velocity, but I don't have a strong opinion.

Makes sense. I guess it mostly just looks a bit strange to write `velocity_rescaling = 'velocity'` but it does make conceptual sense in terms of "NAC vector" and "velocity vector". 

> Thus, I would go with velocity and nac_then_velocity, but I don't have a strong opinion.

Let's go with your suggestion and if we find something better we can change it before the next release.

src/potentials_sh.F90
```diff
@@ -125,7 +125,6 @@ subroutine force_nai(x, y, z, fx, fy, fz, eclas)
       E2 = (VA + VX) / 2.0D0 + dsqrt((VA - VX)**2.0D0 + 4.0D0 * VXA**2.0D0) / 2.0D0
       ! nonadiabatic coupling vector in the reduced system
       d12 = -(VXA * (dVA - dVX) + (-VA + VX) * dVXA) / (VA**2.0D0 - 2.0D0 * VA * VX + VX**2.0D0 + 4.0D0 * VXA**2.0D0)
-      d12 = d12 / ANG ! one more conversion necessary
```
danielhollas, line 128: If this is a bugfix to the NaI potential, can you make a separate PR?

JanosJiri: Kind of a bug fix. A bit of complicated history behind it. Regarding the new PR, the current test of BA uses this bug fix. I would need to do the PR, merge it and then rebase this branch. What if I instead make an issue about this bug and link it to this PR. This would document the bug fix without a new PR. A little bit less work. But if you want a new PR, I can make it. It's up to you @danielhollas .

danielhollas: Yeah, I would prefer to make a new PR, link it to the issue which explains the problem, and merge it first, that should be quick. When it is merged to master you don't need to rebase (I agree that would be annoying), you can simply merge master into this branch. Does that make sense?

It's probably me being picky now, but in general I really don't like when bug fixes like this are hidden in big PRs, it is very hard to track down when you try to understand something a year later.

danielhollas: Yeah, I would prefer to make a new PR, link it to the issue which explains the problem, and merge it first, that should be quick. When it is merged to master you don't need to rebase (I agree that would be annoying), you can simply merge master into this branch. Does that make sense?

It's probably me being picky now, but in general I really don't like when bug fixes are hidden in big PRs, it is very hard to track down when you try to understand something a year later.

sample_inputs/input.in.sh
JanosJiri: > Should we target end of the year for new release? :-) I'll create a separate issue for discussion.

Sounds good, I'm up for that.

src/potentials_sh.F90
JanosJiri: Ok, I created a separate PR with the bug fix.

JanosJiri:
With the last commit and by merging #194 , all the issues raised during the review are resolved. Once you confirm @danielhollas you agree with merging, I will do so.

danielhollas:
@JanosJiri apologies for stalling this, I wanted to have a closer look at the implementation but did not yet find the time.

I am a bit overwhelmed with other stuff this week so I might only get to this next week. If you want this merged go ahead and we can follow-up with a new PR if needed. Thanks for all the cool work on this!

src/surfacehop.F90
danielhollas: Just a note for myself: I don't know why I put the `sh` namelist at the module level, since it is not public, it would be better to put it into `sh_init` function, to better separate variables that are only needed there (such as the new `couplings` keyword.

src/surfacehop.F90
```diff
@@ -227,18 +242,45 @@ subroutine check_sh_parameters()
          error = .true.
       end if
 
-      if (inac > 2 .or. inac < 0) then
-         write (stderr, '(A)') 'Parameter "inac" must be 0, 1 or 2.'
+      ! converting input 'couplings' into inac which is used in the code
+      select case (couplings)
+      case ('analytic')
+         inac = 0
+         write (stdout, '(A)') 'Using analytic ab initio couplings.'
```
danielhollas, line 249: Nice, thanks for including these messages to stdout!

src/surfacehop.F90
```diff
@@ -1012,6 +1108,42 @@ subroutine interpolate(vx, vy, vz, vx_old, vy_old, vz_old, vx_int, vy_int, vz_in
 
    end subroutine interpolate
 
+   ! interpolation of time-derivative coupling calculated via Baeck-An approximation
+   ! this routine interpolates sigma_ba between integration steps
+   subroutine interpolate_ba(vx, vy, vz, vx_old, vy_old, vz_old, vx_int, vy_int, vz_int, &
```
danielhollas, line 1113: I am not stoked about yet another interpolation routine. At some point it would be good to refactor them into a more general routine that could then be called on various arrays. But not in this PR anyway.

src/surfacehop.F90
```diff
@@ -645,15 +730,26 @@ subroutine surfacehop(x, y, z, vx, vy, vz, vx_old, vy_old, vz_old, dt, eclas)
          pop0 = get_elpop(ist)
 
          ! INTERPOLATION
-         fr = real(itp, DP) / real(substep, DP)
-         call interpolate(vx, vy, vz, vx_old, vy_old, vz_old, vx_newint, vy_newint, vz_newint, &
-                          nacx_newint, nacy_newint, nacz_newint, en_array_newint, &
-                          dotproduct_newint, fr)
+         if ((inac == 0) .or. (inac == 2)) then
```
danielhollas, line 733: It's not clear to me if we need to interpolate if inac==2? But better keep the existing behaviour for now.

src/surfacehop.F90
```diff
@@ -568,9 +636,21 @@ subroutine move_vars(vx, vy, vz, vx_old, vy_old, vz_old)
...
+      ! Shift the energy history for Baeck-An couplings
+      if (inac == 1) then
+         ! Move old energies by 1
+         en_hist_array(:, 4) = en_hist_array(:, 3)
+         en_hist_array(:, 3) = en_hist_array(:, 2)
+         en_hist_array(:, 2) = en_hist_array(:, 1)
+         ! new energy is stored before the couplings are calcualted
+         ! I avoided doing the same as with LZSH energy history tracking because then I need to modify force_abin, force_terash and
+         ! every potential in potentials_sh (and all possible future ones). This way it is kept private and does not depend on the
+         ! way energies are calcualted.
```
danielhollas, line 650:
```suggestion
         ! way energies are calculated.
```

src/surfacehop.F90
```diff
@@ -568,9 +636,21 @@ subroutine move_vars(vx, vy, vz, vx_old, vy_old, vz_old)
...
-
       end do
 
+      ! Shift the energy history for Baeck-An couplings
+      if (inac == 1) then
+         ! Move old energies by 1
+         en_hist_array(:, 4) = en_hist_array(:, 3)
+         en_hist_array(:, 3) = en_hist_array(:, 2)
+         en_hist_array(:, 2) = en_hist_array(:, 1)
+         ! new energy is stored before the couplings are calcualted
```
danielhollas, line 647: Thanks for the explanation!
```suggestion
         ! new energy is stored before the couplings are calculated
```

src/surfacehop.F90
```diff
@@ -549,6 +591,32 @@ subroutine calc_nacm(pot, nac_accu)
       end if
    end subroutine calc_nacm
 
+   ! calculate Baeck-An couplings
+   subroutine calc_baeckan(dt)
```
danielhollas, line 595: Would you mind adding a bigger comment about what the Baek-An couplings are? Ideally including one or more references.

src/surfacehop.F90
```diff
@@ -549,6 +591,32 @@ subroutine calc_nacm(pot, nac_accu)
...
+
+      sigma_ba = 0.0D0
+
+      ! we don't have sufficient history until step 4
+      if (it < 4) return
+
+      do ist1 = 1, nstate
+         do ist2 = ist1 + 1, nstate
+            de = en_hist_array(ist2, :) - en_hist_array(ist1, :)
+            de2dt2 = (2.0D0 * de(1) - 5.0D0 * de(2) + 4.0D0 * de(3) - de(4)) / dt**2
```
danielhollas, line 609: Would you mind adding a comment about what this expression is? Perhaps referencing an equation from a paper?
(sorry, I know I am holding this new code to a higher standard than the pre-existing code :see_no_evil:

src/surfacehop.F90
JanosJiri: References added.

src/surfacehop.F90
JanosJiri: Done

src/surfacehop.F90
JanosJiri: I don't think it is necessary to interpolate but the original code was interpolating also for `inac==2` so I kept it because I didn't want to mess up something.

JanosJiri:
> @JanosJiri apologies for stalling this, I wanted to have a closer look at the implementation but did not yet find the time.
>
> I am a bit overwhelmed with other stuff this week so I might only get to this next week. If you want this merged go ahead and we can follow-up with a new PR if needed. Thanks for all the cool work on this!

Ok, thanks for the message and the comments. I will merge it (we need Baeck-An for a project here in Prague) and if you find any problems with implementation, we can fix it in other PR.

danielhollas:
Sounds good! Thanks for the final tweaks and documentation.

PR #187, opened 2024-11-12
CI: Add separate debug build with GCC 13

danielhollas:
Hoping to make things less confusing by having a separate job that has `-check=all` and other runtime  check flags.


JanosJiri (approved):
This looks less confusing, thanks!

PR #190, opened 2024-11-27
Fixes from Fortitude linter

### LINKED ISSUE #189: Investigate using Fortitude: Fortran linter
[Fortitude](https://github.com/PlasmaFAIR/fortitude) is a (very) new Fortran linter inspired by the Python linter Ruff (and actually using some of its code!). 

https://fortran-lang.discourse.group/t/fortitude-a-fortran-linter/8834

It's still very early but just quickly playing with it already proved useful.

Ultimately it would be nice to have it run automatically in CI, once they improve the configuration a bit (for example we need to ignore some of the files)

danielhollas:
Applies fixes from Fortitude linter, see #189 for more info.

You can install fortitude with uv (or pip/pipx)
```console
uv tool install fortitude-lint
```

And run it on the codebase

```console
fortitude check src/
```

Most of the fixes are adding the `intent` attribute to functions/subroutine parameters, missing (or superfluous) `implicit none`, and missing kind parameter on number constants (`0.5` -> `0.5D0`)

In this PR I don't add Fortitude in CI (yet) since it is currently missing some configuration capability that we need. In particular, it's not possible to skip files. We need to skip `FFTW3.F90` and `force_cp2k.F90`. The former is autogenerated file that we should not touch. The latter for some reason contains some weird errors that I was too lazy to understand (and this file is not really important now).


JanosJiri (approved):
All the changes look reasonable.

PR #191, opened 2024-11-27
Rename r.abin to run.abin.sh

### LINKED ISSUE #146: Rename r.abin* scripts
Especially for new users, `r.abin` is not a very good name. We should make clear that it is a BASH script for submitting ABIN job to a scheduler.

Proposed names:
- `submit.abin.sh`
- `submit_abin.sh`

### LINKED ISSUE #196: Minor bug in r.abin
I found a minor bur in [r.abin](https://github.com/PHOTOX/ABIN/blob/master/utils/r.abin) file, specifically lines 13 and 14:

```
#$ -cwd -notify
#$ -q kq-32-8 -pe shm 1
```

#$ is read by Unix systems as a "command" similary to #! when writing #!/bin/bash instead of comment line.

I would recommend changing to:

```
# #$ -cwd -notify
# #$ -q kq-32-8 -pe shm 1
```

danielhollas:
I would like to rename `r.abin` and similar scripts, since the name can be a bit weird to newcommers. For now I decided for `run.abin.sh`, but I would love to hear feedback on the naming!

Closes #146.
Closes #196


JanosJiri (approved):
Renaming `r.abin` seems like a good idea. I would maybe use `run.terachem_mpi_abin.sh` instead of `run.tc_mpi_abin.sh`. Not everyone will know that tc means terachem.

PR #201, opened 2025-03-05
Update fortitude linter to v0.7

danielhollas:
A new version of fortitude linter has been released, with lot's more configuration options:

https://github.com/PlasmaFAIR/fortitude/releases/tag/v0.7.0

This will allow us to run it in CI, I'll add it in a follow-up PR.

PR #202, opened 2025-03-05
CI: Add fortitude linter check

danielhollas:
This adds a new CI jobs that runs the fortitude linter. Depends on #201


danielhollas:
Ugh, GitHub removed Ubuntu 20.04 runners so we need to upgrade (which means dropping testing for GCC-7)

PR #208, opened 2025-07-15
Interfaces to ChemShell

postulkj:
Adding ChemShell interface for QM/MM calculation. Both standard and LZ version.

### DIFF
--- a/.github/workflows/format-lint.yml
+++ b/.github/workflows/format-lint.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-24.04
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
     - uses: astral-sh/setup-uv@v6
 
     - name: Run formatter
@@ -27,7 +27,7 @@ jobs:
     name: Lint check
     runs-on: ubuntu-24.04
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
 
     - name: Install fortitude linter
       run: pip install fortitude-lint==0.7.1
@@ -39,7 +39,7 @@ jobs:
     name: Test and lint utililities and analysis scripts
     runs-on: ubuntu-24.04
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
     - uses: astral-sh/setup-uv@v6
     - name: Run scripts via uv run 
       run: uv run -p 3.9 utils/analysis/tsh_traj_anal.py -h
--- a/.github/workflows/gfortran.yml
+++ b/.github/workflows/gfortran.yml
@@ -45,7 +45,7 @@ jobs:
       CODECOV_NAME: ${{format('{0} GCC-{1}', github.job, matrix.gcc_v)}}
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
       with:
         fetch-depth: 2
 
@@ -98,7 +98,7 @@ jobs:
     runs-on: ubuntu-24.04
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
 
     - uses: fortran-lang/setup-fortran@v1
       id: setup-fortran
@@ -120,7 +120,7 @@ jobs:
     runs-on: ubuntu-22.04
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
 
     - uses: fortran-lang/setup-fortran@v1
       id: setup-fortran
@@ -144,7 +144,7 @@ jobs:
       FC: gfortran
       CODECOV_NAME: ${{format('{0}', github.job)}}
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
       with:
         fetch-depth: 2
 
@@ -192,7 +192,7 @@ jobs:
       GCC_V: ${{ matrix.gcc_v}}
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
       with:
         fetch-depth: 2
     - name: Set GFortran version
@@ -231,7 +231,7 @@ jobs:
     env:
       CODECOV_NAME: ${{format('{0}', github.job)}}
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
       with:
         fetch-depth: 2
 
@@ -291,7 +291,7 @@ jobs:
       CODECOV_NAME: ${{format('{0} GCC-{1} MPICH-{2}', github.job, matrix.gcc_v, matrix.mpich_v)}}
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
       with:
         fetch-depth: 2
 
@@ -351,7 +351,7 @@ jobs:
       OPENMPI_PATCH: "2"
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
       with:
         fetch-depth: 2
     - name: OpenMPI build cache
@@ -395,7 +395,7 @@ jobs:
       CODECOV_NAME: ${{format('{0} PLUMED-{1}', github.job, matrix.plumed_v)}}
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
       with:
         fetch-depth: 2
     - name: Plumed build cache


interfaces/CHEMSH-LZ/r.chemsh-lz
```diff
@@ -0,0 +1,278 @@
+#!/bin/bash
+
+#This script is used to run QM/MM calculations with ChemShell with Landau-Zener surface hopping.
+#Requred qm programs like MOLPRO have to be source to run it in chemshell environment.
+
+#ChemShell
+source SetEnvironment.sh CHEMSHELL
```
JanosJiri, line 7: This is purely a PHOTOX feature. It would be good to explain what this is supposed to do.

interfaces/CHEMSH/r.chemsh
```diff
@@ -0,0 +1,239 @@
+#!/bin/bash
+
+#This script is used to run QM/MM calculations with ChemShell.
+#Requred qm programs like MNDO or ORCA have to be source to run it in chemshell environment.
+
+#ChemShell
+source SetEnvironment.sh CHEMSHELL
+
+#QM Progam
+source SetEnvironment.sh MNDO
```
JanosJiri, line 10: Really sourcing two QM codes at the same time?

interfaces/CHEMSH/r.chemsh
```diff
@@ -0,0 +1,239 @@
...
+
+cd CHEMSH
+timestep=$1
+ibead=$2
+input=input$ibead
+natom=$(wc -l ../geom.dat.$ibead) && natom=$(echo "${natom%% *}")
+geometry=../geom.dat.$ibead
+
+
+#################### USER INPUT #######################
```
JanosJiri, line 21: Again the same comments as for the previous file.

interfaces/CHEMSH/r.chemsh
```diff
@@ -0,0 +1,239 @@
...
+QMatom=24            # Number of QM atoms
+hamiltonian="mndo"      # QM method
```
JanosJiri, line 23: Explaining options? The probably sourcing the QM code based on the option.

interfaces/CHEMSH/r.chemsh
```diff
@@ -0,0 +1,239 @@
...
+basis="cc-pVDZ"
```
JanosJiri, line 24: Not used at all. I also don't see where the input for the QM program is built? How do you get the QM input?

JanosJiri (changes requested):
Great work bringing a new interface to ABIN. I have some comments to address though as I didn't get some things.

interfaces/CHEMSH-LZ/r.chemsh-lz
```diff
@@ -0,0 +1,278 @@
...
+cd CHEMSH-LZ
+timestep=$1
+ibead=$2
+input=input$ibead
+natom=$(wc -l ../geom.dat.$ibead) && natom=$(echo "${natom%% *}")
+geometry=../geom.dat.$ibead
+read -t 2 -a nstate
+read -t 2 -a tocalc
+
+#################### USER INPUT #######################
```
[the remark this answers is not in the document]
postulkj: I hate it, but OK. It is done.

interfaces/CHEMSH-LZ/r.chemsh-lz
```diff
@@ -0,0 +1,278 @@
...
+QMatom=4            # Number of QM atoms
+nelectrons=18  # total number of electrons
+pspace=1
+spin=0       # 0 for singlet, 1 for dublet etc.
+nocc=11      # occupied orbitals
+nclosed=6     # closed orbitals
+memory=600     # molpro memory in MegaWords (1MW = 8 Mb)
+method="casscf"
```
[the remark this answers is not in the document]
postulkj: nocc, nclosed and memory are used to create 'molpro.com'. method was not used. It is now solved and commented so the reader knows to provide their own molpro.com (or use default CASSCF).

interfaces/CHEMSH-LZ/r.chemsh-lz
```diff
@@ -0,0 +1,278 @@
...
+# Every row represents one atom type and its epsilon and sigma parameters.
+# Here units of kJ/mol and nm are expected.
+# These parameters will be transformed and added to dl_poly input file ff.dat. Pair rules will be applied on them and on those found in ff.dat.
+# THIS INTERFACE HAS AN EXAMPLE OF HYDROGEN PEROXIDE IN FLEXIBLE SPC WATER FROM  https://doi.org/10.1063/1.2136877
+cat > non_bonded.dat << EOF
+ox 2.4058  0.2912
+hx 0.527184 0.1852
+EOF
+
+#################### END OF USER INPUT ################
```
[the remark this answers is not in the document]
postulkj: Done

interfaces/CHEMSH-LZ/r.chemsh-lz
```diff
@@ -0,0 +1,278 @@
...
+echo '' >> input.xyz
+cat $geometry >> input.xyz
+geometry=input.xyz
+MMatom=$( echo "$natom-$QMatom" | bc )
+length_residue=$(echo $types_mm_atoms | wc -w)
+nresidue=$(echo $MMatom / $length_residue | bc)
+
+
+############### MOlPRO INPUT ####################
+if [ ! -f molpro.com ]; then
```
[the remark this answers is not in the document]
postulkj: Good point. Done.

interfaces/CHEMSH/r.chemsh
postulkj: Changed by adding new keyword qm_code and if block for sourcing only one.

interfaces/CHEMSH/r.chemsh
postulkj: The QM input is build by chemshell. The interface just provides method and other parameters through chemshell keyword 'qm_theory'. In this case the basis is not used, but it should be defined in case of orca.

interfaces/CHEMSH-LZ/r.chemsh-lz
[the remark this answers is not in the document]
danielhollas: Just for the context, the splitting in terms of SH interfaces was done mainly due to there being two different files already, one for the energies and forces and second one for NA couplings, so having the input parameters in separate file was important for consistency. 

But in general I do think it's nice to clearly have a file that should be modified and one that should not be touched. But it is currently not consistent, and not a huge deal.

interfaces/CHEMSH/r.chemsh
```diff
@@ -0,0 +1,206 @@
...
+cat conn.dat >> geom.c
+fi
+#################### Launch ChemShell ####################
+
+$CHEMSHELLEXE $input.com > $input.com.out
+
+if [[ $? -eq 1 ]];then
+   echo "Error in ChemShell. Check $input.com.out"
+elif $( grep -q grep 'error code 1' $input.com.out ) ;then
+   echo "Error in QM code."
```
danielhollas, line 154:
```suggestion
if [[ $? -ne 0 ]];then
   echo "Error in ChemShell. Check $input.com.out"
   exit 2
elif $( grep -q grep 'error code 1' $input.com.out ) ;then
   echo "Error in QM code."
   exit 2
```

interfaces/CHEMSH-LZ/r.chemsh-lz
```diff
@@ -0,0 +1,238 @@
...
+cat conn.dat >> geom.c
+fi
+#################### Launch ChemShell ####################
+
+$CHEMSHELLEXE $input.com > $input.com.out
+
+if [[ $? -eq 1 ]];then
+   echo "Error in ChemShell. Check $input.com.out"
+elif $( grep -q grep 'error code 1' $input.com.out ) ;then
+   echo "Error in QM code."
```
danielhollas, line 182: Any non-zero exit code should be treated as an error, and also you want to exit early when there's an error

```suggestion
if [[ $? -ne 0 ]]; then
   echo "Error in ChemShell. Check $input.com.out"
   exit 2
elif $( grep -q grep 'error code 1' $input.com.out ) ;then
   echo "Error in QM code."
   exit 2
```

Also the error code of chemshell invocation on line 148 should be checked as well I think

danielhollas (changes requested):
This is awesome to see, impressive and non-trivial work! :clap: 

Just one suggestion with regards to catching errors.

Overall I think here you've almost reached the limit of practicality of bash scripts.
At this point I'd consider a more structured python script as an interface (but we currently don't have any python interfaces to start from, and Python has non-trivial startup cost (10-20ms), but that probably would not matter in this case.
(not saying anything should be changed here, just thinking out loud).

postulkj:
Thanks for all the suggestions. Everything should be resolved now.

danielhollas (approved):
Amazing work, thank you!

PR #224, opened 2026-03-07
feat: add MACE interface

FelyCZ:
# Add MACE Interface via MPI

## Description

This PR adds a new interface for the [MACE](https://github.com/ACEsuit/mace) (Machine Learning Atomic Cluster Expansion) potential to ABIN. MACE is integrated via an MPI-based client-server architecture: a Python server loads and evaluates the MACE ML model, while the Fortran side communicates coordinates, energies, and forces over MPI. This follows the same pattern used for the existing TeraChem (`_tera_`) interface.

The interface supports configurable model paths, device selection (CPU/CUDA), precision settings, and various MACE-specific options -- all controlled through a new `&mace` namelist in the ABIN input file.

---

### Fortran Core

| File | Description |
|------|-------------|
| **`src/mace_mpi_api.F90`** | **[NEW]** Core MPI API module (`mod_mace_mpi`). Handles MPI port connection to the Python server, sends atom types/config/coordinates, and defines all MACE namelist variables (model path, device, dtype, batch size, etc.). |
| **`src/force_mace.F90`** | **[NEW]** Force driver module (`mod_force_mace`). Orchestrates per-bead force evaluation -- sends coordinates via `send_mace`, receives energy and forces via `receive_mace`. Supports OpenMP. |
| **`src/forces.F90`** | Added `_mace_` case to the `force_wrapper` potential selector, routing to `force_mace()`. |
| **`src/init.F90`** | Added `&mace` namelist reading and MACE interface/server initialization when `pot='_mace_'`. |
| **`src/mpi_wrapper.F90`** | Added `_mace_` to the MPI-required potential check so non-MPI builds raise a clear error. |
| **`src/Makefile`** | Added `mace_mpi_api.o` and `force_mace.o` to the build. |

### Python Server

| File | Description |
|------|-------------|
| **`interfaces/MACE/mace_server.py`** | **[NEW]** MACE MPI server. Opens MPI port, accepts connection from ABIN, receives configuration and atom data, loads the MACE model, and returns energies/forces each step. Handles unit conversions (Bohr <-> Å, Hartree <-> eV) and supports foundation model auto-download. |
| **`interfaces/MACE/README.md`** | **[NEW]** Documentation covering requirements, manual/automatic launch instructions, and a full configuration reference table for all `&mace` namelist parameters. Includes MACE-OFF licensing notice. |
| **`interfaces/MACE/MACE-OFF23_medium.model`** | **[NEW]** Bundled pre-trained MACE-OFF23 medium model (~18 MB) for out-of-the-box usage. Licensing notice is included in the README. |

### Tests

| File | Description |
|------|-------------|
| **`tests/MACE/test.sh`** | **[NEW]** Integration test script. Launches the mock MACE server and ABIN via MPI, runs a short MD simulation, and validates output against reference files. |
| **`tests/MACE/mock_mace_server.py`** | **[NEW]** Mock MACE server for testing. Implements the same MPI protocol as the real server but returns forces from a simple harmonic potential, avoiding the need for `mace-torch`/PyTorch. |
| **`tests/MACE/input.in`** | **[NEW]** Test input file with `pot='_mace_'` and `&mace` namelist. |
| **`tests/MACE/mini.xyz`** | **[NEW]** Test geometry (water molecule). |
| **`tests/MACE/*.ref`** | **[NEW]** Reference output files (`energies.dat.ref`, `forces.xyz.ref`, `movie.xyz.ref`, `velocities.xyz.ref`, `temper.dat.ref`). |
| **`tests/test.sh`** | Registered `MACE` in the MPI test folder list. |
| **`unit_tests/test_mace.pf`** | **[NEW]** pFUnit tests verifying default namelist values, MPI tag constants, and non-MPI error handling. |
| **`unit_tests/Makefile`** | Added `mace` target to the unit test build, test runner, and clean targets. |

### Utilities & Docs

| File | Description |
|------|-------------|
| **`utils/run.mace_mpi_abin.sh`** | **[NEW]** Launch script that starts both the MACE Python server and ABIN, connects them via MPI, and monitors both processes with automatic cleanup. |
| **`dev_scripts/install_mace.sh`** | **[NEW]** Convenience script to clone the MACE repo and install dependencies (`mace-torch`, `mpi4py`). |
| **`sample_inputs/input.in.mace`** | **[NEW]** Sample input file demonstrating `&mace` namelist configuration. |
| **`README.md`** | Added MACE to the list of optional libraries with a brief description and install reference. |


interfaces/MACE/MACE-OFF23_medium.model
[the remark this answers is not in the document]
FelyCZ: I agree that it's not entirely necessary. The main idea was to provide users with a quick way to test the interface without any further configuration.

The file is not necessary for the tests; however, the default configuration (`&mace` namelist) uses the file as fallback.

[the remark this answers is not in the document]
FelyCZ: Hi, @danielhollas,
Are you still of the opinion that we should remove the model from the repo?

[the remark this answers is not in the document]
danielhollas: Hi, yes I think we should remove it. Please add instructions somewhere (perhaps a README file) where we can tell the users how to download it.

Sorry for the slow review, this is a big PR and I want to have a careful look and we also plan to test it soon.

dev_scripts/install_mace.sh
```diff
@@ -0,0 +1,28 @@
...
+fi
+
+if [[ -e $REPO_DIR ]];then
+  echo "ERROR: $REPO_DIR already exists."
+  exit 1
+fi
+
+git clone "${REPO_URL}" "${REPO_DIR}" && cd "$REPO_DIR"
+
+pip install --upgrade pip
```
danielhollas, line 17: This will install the package into the global python environment (unless the user is already inside a venv or conda. I think we should create a fresh venv for the user.

dev_scripts/install_mace.sh
```diff
@@ -0,0 +1,28 @@
...
+pip install .
```
danielhollas, line 18: Why are we installing MACE from the repo? We should I think install the package from PyPI.

dev_scripts/install_mace.sh
```diff
@@ -0,0 +1,28 @@
...
+pip install mpi4py
```
danielhollas, line 19: To be clear, I think we should do something like this:

```
python -m venv .mace_venv
.mace_venv/bin/python3 -m pip install mace-torch mpi4py
```

interfaces/MACE/mace_server.py
```diff
@@ -0,0 +1,319 @@
...
+        self.return_contributions = config.get('return_contributions', 'false').lower() == 'true'
+        self.info_prefix = config.get('info_prefix', 'MACE_')
+        head_val = config.get('head', '')
+        self.head = head_val if head_val else None
+
+        torch_tools.set_default_dtype(default_dtype)
+        self.device = str(torch_tools.init_device(device))
+
+        import os
+        if os.path.isfile(model_path):
```
danielhollas, line 91: This interface is a bit confusing. Should we instead have two diferrent (mutually exclusive) parameters, `model_path` and `model_name`?

interfaces/MACE/mace_server.py
```diff
@@ -0,0 +1,319 @@
...
+            # Load from local file
+            try:
+                self.model = torch.jit.load(f=model_path, map_location=device).to(device)
```
danielhollas, line 94: I am a little confused as to why we're importing the model directly to torch, and not using MACE specific functions / interface for it?

interfaces/MACE/mace_server.py
```diff
@@ -0,0 +1,319 @@
...
+          - medium, small, large (shorthand for MACE-OFF23)
+          - MACE-MP-0_medium.model, etc.
+        """
+        from mace.calculators.foundations_models import mace_off, mace_mp
+
+        name = model_name.replace('.model', '').strip()
+        log(f"Loading foundation model: {name}")
+
+        # Parse model family and size
+        name_lower = name.lower()
```
danielhollas, line 121: To simplify the if-statements:


```suggestion
        name_lower = name.lower().replace('-', '_')
```

interfaces/MACE/mace_server.py
```diff
@@ -0,0 +1,319 @@
...
+        else:
+            # Try loading as a MACE foundation model (auto-downloads)
+            self.model = self._load_foundation_model(model_path, device)
+
+        for param in self.model.parameters():
+            param.requires_grad = False
+
+        log("model loaded in %.2f s" % (time.time() - start_time))
+
+    def _load_foundation_model(self, model_name, device):
```
danielhollas, line 107: This interface seems brittle. Why not accept a model name (in some canonical form), and have a size parameter separately?

```suggestion
    def _load_foundation_model(self, model_name, size, device):
```

interfaces/MACE/mace_server.py
```diff
@@ -0,0 +1,319 @@
...
+            calc = mace_mp(model=size, device=device, return_raw_model=True)
+        else:
+            raise RuntimeError(
+                f"Unknown model '{model_name}'. Provide a path to a local .model file, "
+                f"or use a foundation model name like 'MACE-OFF23_medium.model' or 'medium'."
+            )
+
+        return calc
+
+    def evaluate(self, atom_types, coords_bohr):
```
danielhollas, line 143: Here I am again confused. Why are we not simply calling a MACE calculator here?

interfaces/MACE/mace_server.py
```diff
@@ -0,0 +1,319 @@
...
+
+
+def main():
+    comm = MPI.COMM_WORLD
+
+    # Open MPI port and write to file for ABIN to read
+    port_name = MPI.Open_port()
+    log(f"MPI port opened: {port_name}")
+
+    port_file = "mace_port.txt.1"
```
danielhollas, line 223: Let's define this as a "constant" at the top of the script, something like:

```python
MACE_PORT_FILE = "mace_port.txt.1"
```

interfaces/MACE/README.md
```diff
@@ -0,0 +1,73 @@
...
+## Usage
+
+The MACE server is typically launched alongside ABIN using MPI. ABIN communicates with the server to obtain energies and forces.
+
+### Manual Launch (OpenMPI)
+
+When using OpenMPI, you need an `ompi-server` running for the connection handshake:
+
+```console
+# 1. Start ompi-server
```
danielhollas, line 32: Is this really the case? We don't seem to need something like this in the TeraChem interface?

interfaces/MACE/README.md
```diff
@@ -0,0 +1,73 @@
...
+### Automatic Launch
+
+It is recommended to use the provided launch script in `utils/`:
+```console
+./utils/run.mace_mpi_abin.sh
+```
+
+## Configuration
+
+MACE settings are controlled via the `&mace` namelist in the ABIN input file (`input.in`). Below is the full list of available parameters:
```
danielhollas, line 51: So I am not totally opposed to this, but note that generally we don't put interface-specific parameters in the abin input file since ABIN doesn't really need to know about these.

 I feel that in this case it would be better (and more general!) to hava a YAML configuration file in the interface folder which we can than load directly into MACE (I believe MACE supports this natively).

src/init.F90
```diff
@@ -149,6 +150,10 @@ subroutine init(dt)
 
       namelist /qmmm/ natqm, natmm, q, LJ_rmin, LJ_eps, mm_types
 
+      namelist /mace/ mace_model, mace_device, mace_default_dtype, &
```
danielhollas, line 153: Yeah, I don't really think we should be parsing MACE specific keywords here, they are irrelevant from ABIN's point of view (and note that since MACE excosystem is still evolving a lot, we need to have the flexibility to add new options easily, without having to add them here and recompile ABIN

src/mace_mpi_api.F90
```diff
@@ -0,0 +1,318 @@
...
+         write (stderr, *) get_mpi_error_string(ierr)
+      end if
+
+      call MPI_Comm_free(mace_comm, ierr)
+      if (ierr /= MPI_SUCCESS) then
+         write (stderr, *) get_mpi_error_string(ierr)
+      end if
+   end subroutine finalize_mace
+
+   subroutine wait_for_mace(comm)
```
danielhollas, line 170: This code was added to TeraChem interface because a typing TC calculation takes (much) more than a 1 second so we could save a whole CPU by instructing ABIN to sleep a bit while waiting for TC.

 In MACE case, the hope is that it is actually very quick, so I don't think this case makes sense. I'd just remove it for now to keep things simple? (happy to hear otherwise though!)

danielhollas (changes requested):
Did a very quick initial review, I hope these comments are useful.

dev_scripts/install_mace.sh
danielhollas, line 1: Thinking about this more, I don't think we should provide the `install_mace.sh` script, there's simply too many variables about how to handle the python environment (and how to handle Torch installation in particular).

I'd rather have a comprehensive installation section in the README. Given the (unfortunate) fact that the `mace` package is not published on conda, I think we should encourage users to use the `uv` package installer, in particular because it has an advanced handling of pytorch.

https://docs.astral.sh/uv/guides/integration/pytorch/

i.e. something like this

```
uv venv
uv pip install mace-torch mpi4py --torch-backend=auto
```

To install CPU-only version

```
uv pip install mace-torch mpi4py --torch-backend=cpu
```


I am happy to help write this down before we merge this PR.

interfaces/MACE/mace_server.py
FelyCZ: We didn't use the default pre-trained model; we used a custom, fine-tuned model based on MACE-OFF-23. However, the MACE functions do not support the use of custom models (afaik). Therefore the interface is based on [eval_configs.py](https://github.com/ACEsuit/mace/blob/main/mace/cli/eval_configs.py) file.

danielhollas: Can you say a bit more about your fine-tuned model? Is it still in a `*.model` format? 
The `mace.calculators.MACECalculator` definitely supports arbitrary models, not just the pre-trained ones. See for example the code here:

https://github.com/duartegroup/mlp-train/blob/64d2b0010d40e63b733febfb57679ad2f2e90601/mlptrain/potentials/mace/mace.py#L242

danielhollas:
Hi @FelyCZ, just wanted to ask if you'll have time to work on this further in the near future?(No worries if not)

I am hoping to do some testing on this branch next week, as we will need this interface soon in our own MACE project.

FelyCZ:
Hi @danielhollas  , 

I am very sorry but I have so many things with deadlines piling up now. I probably won't be able to work on this at least till the end of May.

danielhollas:
@FelyCZ no problem, thanks for letting me know! 

I obviously don't want to take over your great work, but also since we need to press on with our project, I am wondering what would be best for now, I see couple of options:

1. We merge this PR more or less as is, to get it out of your TODO list and to have you have the full credit for the feature. I could then continue building on top of it.
2. If you'd like to go through the review yourself and have full ownership, I could in the meantime open a separate branch based on this one so I can do any necessary changes to unblock our work. And then once you have time we'd continue working together on this PR and merge it once it's fully polished. 
3. Something else?

Let me know what would you prefer, I am fine either way. Also let me know if there's anybody else in Petr's group that would be interested in working on this (I know Jirka is busy, perhaps somebody else?)

danielhollas:
I pushed a few commits and the MACE test now passes on the OpenMPI build, very cool! :tada: MPICH build still fails, I'll have a look a bit more later.

FelyCZ:
> @FelyCZ no problem, thanks for letting me know!
> 
> I obviously don't want to take over your great work, but also since we need to press on with our project, I am wondering what would be best for now, I see couple of options:
> 
> 1. We merge this PR more or less as is, to get it out of your TODO list and to have you have the full credit for the feature. I could then continue building on top of it.
> 2. If you'd like to go through the review yourself and have full ownership, I could in the meantime open a separate branch based on this one so I can do any necessary changes to unblock our work. And then once you have time we'd continue working together on this PR and merge it once it's fully polished.
> 3. Something else?
> 
> Let me know what would you prefer, I am fine either way. Also let me know if there's anybody else in Petr's group that would be interested in working on this (I know Jirka is busy, perhaps somebody else?)

Hi @danielhollas ,
Sorry for the delayed response. I am fine with either of the options listed. I see that you did quite a bit of work related to the build.
Did you end up continuing with the feature on your own? I might be able to squeeze in a little time to resolve some of the comments in the next week or two.

Regarding the other person from Photox who would be interested, Martin Prochazka might be a good candidate since he recently messaged me about the status of this. I will reach out to him.

danielhollas:
@FelyCZ great, thanks for the info! Yes, definitely feel free to go ahead and work on this branch. Here as you noticed I mostly just focused on making the CI green. 

I also have a separate branch based on this with couple more fixes, most notably today we fixed an issue with the float32 precision. I am hoping with that fix in place we might be off to actually run production MD! Feel free to take a look at the branch, I haven't changed much and not things that I've pointed out in my initial review, so definitely it would be great if you can spend some time on this. https://github.com/PHOTOX/ABIN/tree/dh-mace

Also, I will be in Prague for the week starting 15th of June, working with Petr and Jirka on ABIN paper. So that might be a nice opportunity to work on this together in case your interested.

> Regarding the other person from Photox who would be interested, Martin Prochazka might be a good candidate since he recently messaged me about the status of this. I will reach out to him.

Perfect, thanks!

danielhollas:
@FelyCZ so I spoke too soon, and we actually cannot make the interface work, the molecule always blows up, regardless whether we use our own trained model or the MACE-OFF model included in this PR.

Would you mind sending me the exact input files that work for you? (attach here or via email at daniel.hollas@bristol.ac.uk) Also, if you could post an output of `pip list` or `conda list` from your MACE python environment that would be very useful, since we likely run different versions of pytorch / mace etc.

For the record, here are the things I've tried during debugging:

<details><summary>Debugging details</summary>
<p>


- I've tested for both our trained MHP model (float32) and MACE-OFF model (float64) on a single water molecule, both blow up instantly. So this doesn't seem pertinent to our own model, or its precision.
- Tested that blow up happens for both CPU and GPU inference, and tested that batch_size parameter doesn't matter.
- I verified that the energies and forces that are sent from the python mace server are received by ABIN correctly, which exonerates the MPI interface I think.
- Notably, the potential energy goes down when the molecule blows up. Which indicates that this is not some simple unit conversion issue (or an issue with forces having a wrong sign or something). So really seems like the model inference is wrong.
- Also verified that the data that ABIN sends are received and the data that are passed to the inference seem reasonable. (the cell is zero, but PBC is set to False so hopefully that should work)
- Tried to update mace-torch package to latest but did not help.

</p>
</details>

danielhollas:
Turns out the problem was in the way the server was receiving coordinates from ABIN, see the fix in https://github.com/PHOTOX/ABIN/pull/242/commits/f0b0610074fa2d2ea5f6c19a167dfd9baffcd18c

Without this fix I don't see how the interface ever worked?

I've opened #242 with the fix and some other refactor, let's continue development there.

danielhollas:
I'll close this for now and continue in #242

PR #225, opened 2026-03-09
OPENQP Landau-Zener interface

JanosJiri:
Interface to MRSF-TDDFT implemented in OPENQP. Currently only for Landau-Zener SH. 

Created by Katarina Vosovicova (and JJ).


danielhollas (changes requested):
Great! Please see some minor suggestion below. 

The main suggestion I have is to change the approach of generating the openqp input file. Instead of generating it inside the bash script, it would be simpler to have a template input file (e.g. `openqp.inp.template`) and use sed to only modify the parts that are variable during MD (e.g. path to geometry, path to guess, number of states etc). 
The advantage would be that if the user wants to e.g. add some parameter to the input file, they can simply modify just the template file, instead of having to add the variable to the `openqp.inp` and then print it in the script. Does that make sense?

I think SHARK uses a similar strategy, and it's probably something we should do in other interfaces as well in the future.
Note that specifically for Surface Hopping we couldn't do that for Molpro since Molpro's input file is way more complicated and we  do need to generate it on the fly.

JanosJiri:
Thanks for all the suggestions, @danielhollas! Katka processed them all and created this new version. However, I don't understand why the tests fail since we uploaded only an interface; we did not touch the code at all.

danielhollas (approved):
This is great, thanks!

PR #226, opened 2026-03-17
Fix GCC-9 build

danielhollas:


JanosJiri (approved):
Nice, thanks for the fix!

PR #236, opened 2026-04-29
Use prek to run linters and formatter

danielhollas:
Please read the updated CONTRIBUTING.md for the explanation of the changes here and let me know if anything could be clearer!

Here's a rendered CONTRIBUTING.md: https://github.com/PHOTOX/ABIN/blob/489fdc1fe78f724cf39fe52b826dac7defe86515/CONTRIBUTING.md


dev_scripts/pre-push.hook
danielhollas, line 1: Not really needed now

dev_scripts/setup_dev_environment.sh
danielhollas, line 1: not needed anymore. users can run `setup_vim.sh` separately if needed

prek.toml
```diff
@@ -0,0 +1,52 @@
...
+]
+
+[[repos]]
+# shellcheck: A shell linter, https://www.shellcheck.net/
+repo = "https://github.com/koalaman/shellcheck-precommit"
+rev = "v0.11.0"
+hooks = [
+  {
+    id = "shellcheck",
+    files = "dev_scripts/setup_vim.sh"
```
danielhollas, line 35: We should use shellcheck on all the bash scripts, but that's for another PR :-)

prek.toml
danielhollas, line 1: This is basically the same configuration that I am using in OpenFMS now

JanosJiri (approved):
This looks pretty cool!

danielhollas:
@JanosJiri I am curious, can you try running prek hooks on your machine? I suspect the shellcheck hook might fail for you since it requires Docker. Perhaps we should just remove it.

danielhollas:
> @JanosJiri I am curious, can you try running prek hooks on your machine? I suspect the shellcheck hook might fail for you since it requires Docker. Perhaps we should just remove it.

@JanosJiri I am still curious about this :-)

PR #242, opened 2026-05-30
MACE interface, vol. 2

danielhollas:
This is a continuation from #224, with the following changes

- remove `&mace` section from the abin input file, instead the path to the model and other configuration is passed via command line parameter to `mace_server.py`. This is specified at the top of `utils/run.mace_abin.sh`
- Fix coordinates received from ABIN (without this the interface was simply not working)
- Use MACECalculator instead of calling pytorch directly
- Fix float32 evaluation
- removed `install_mace.sh` and the MACE-OFF model file
- test `mace_server.py` in MACE test directly
- lot's of error-handling improvements
- add tests for error paths

TODO: 
 - [x] Test run.mace_abin.sh
 - [x] Let Veronika test everything
 - [x] Check size of incoming coordinates from ABIN in mace_server.py
 - [ ] Use logging module?


danielhollas:
Okay, I think this is mostly ready now for merging! We'll do some internal testing. @FelyCZ would be great if you could have a look. I summarized the changes I made in the PR description.

@JanosJiri up to you how much you want to review this. Perhaps at least the README files and `mace_server.py` for clarity. I'll ask people here for review as well.

JanosJiri (commented):
Hi @danielhollas, I'm sorry but I struggle with time now and I can't make a useful review in a short time. I checked README.md though and it looks clean and understandable to me. Sorry.

interfaces/MACE/mace_server.py
```diff
@@ -0,0 +1,322 @@
...
+        abin_comm.Recv(
+            [coords, MPI.DOUBLE], source=0, tag=MACE_TAG_DATA, status=mpi_status
+        )
+        assert mpi_status.Get_elements(MPI.DOUBLE) == natom * 3
+
+        start = perf_counter()
+
+        energy, forces = mace_model.evaluate(atom_types, coords)
+
+        end = perf_counter()
```
FelyCZ, line 294: I think we need to include `torch.cuda.synchnronize()` when the computation is running on CUDA. Otherwise is won't be measuring the time correctly.

```python
if config.device == "cuda":
    import torch
    torch.cuda.synchronize() 

end = perf_counter()
```

src/force_mace.F90
```diff
@@ -0,0 +1,121 @@
+module mod_force_mace
+! ----------------------------------------------------------------
+! Interface for MACE (Machine Learning Atomic Cluster Expansion)
+! Perform MPI communications with the MACE Python server.
+!
+! Modeled after mod_force_tera for TeraChem.
+! ----------------------------------------------------------------
+   use mod_const, only: DP
+   use mod_error, only: fatal_error
+   use mod_files, only: stdout, stderr
```
FelyCZ, line 10: Move `natqm` to global scope (as it is used in both subroutines)?

```fortran
use mod_qmmm, only: natqm
```

FelyCZ:
Hello, thanks for revising the MACE interface!
I think it is now ready to be released. I have left you some comments and implemented the logging module (see PR #257).

interfaces/MACE/mace_server.py
danielhollas: Is it really needed? Surely `mace_model.evaluate` cannot return the final result until CUDA kernels finished?

danielhollas:
@FelyCZ thanks for taking a look! Let's merge this, and we can continue with tweaks in #257.

We've tested this locally with our own MACE model. We might need to do further testing with various foundational model but that can be a follow-up as well.

Thanks for your original work on this @FelyCZ! I think this is going to be super-useful going forward.

src/force_mace.F90
danielhollas: Yeah, we could, but it also is nice to have it imported locally as it's closer to its use. I'll leave it like that for now.

PR #248, opened 2026-06-17
Velocity rescaling purely along NAC

### LINKED ISSUE #247: Velocity rescaling in SH
Currently, SH offers only `nac_then_velocity` and `velocity` options. We should also add the option `nac` and make it default, which is the standard in the field currently. It is also connected to the `revmom` issue #203, where we should reverse along the NAC vector ideally.

JanosJiri:
Closes #247, see the issue for details.


src/surfacehop.F90
```diff
@@ -999,6 +1010,7 @@ subroutine try_hop_nacme_rescale(vx, vy, vz, instate, outstate, eclas)
       real(DP) :: a_temp, b_temp, c_temp, g_temp
       real(DP) :: ekin, ekin_new
       integer :: iat, iw
+      logical, intent(in) :: isotropic_fallback
```
danielhollas: nit: Generally I like to have all the input parameters defined together

```suggestion
      logical, intent(in) :: isotropic_fallback
      real(DP) :: a_temp, b_temp, c_temp, g_temp
      real(DP) :: ekin, ekin_new
      integer :: iat, iw
```

src/surfacehop.F90
```diff
@@ -1167,6 +1188,8 @@ subroutine try_hop_simple_rescale(vx, vy, vz, instate, outstate, eclas)
       dE = en_array(outstate) - en_array(instate)
       ekin = ekin_v(vx, vy, vz)
 
+      write (*, '(A,E17.10,A,E17.10)') '# deltaE_pot / a.u. = ', dE, ' E_kin-total / a.u. = ', ekin
```
danielhollas:
```suggestion
      write (stdout, '(A,E17.10,A,E17.10)') '# deltaE_pot / a.u. = ', dE, ' E_kin-total / a.u. = ', ekin
```

src/surfacehop.F90
```diff
@@ -1189,21 +1212,28 @@ subroutine try_hop_simple_rescale(vx, vy, vz, instate, outstate, eclas)
...
+         call frustrated_hop(vx, vy, vz, instate, outstate)
 
       end if
 
-      write (*, '(A,E17.10,A,E17.10)') '# deltaE_pot / a.u. = ', dE, ' E_kin-total / a.u. = ', ekin
-
    end subroutine try_hop_simple_rescale
 
+   subroutine frustrated_hop(vx, vy, vz, instate, outstate)
+      use mod_interfaces, only: force_clas
```
danielhollas:
```suggestion
```

danielhollas (approved):
Awesome, thanks, just a couple of small things

PR #250, opened 2026-06-19
Turning off decoherence correction with a specific keyword

### LINKED ISSUE #188: Turning off decoherence correction
The decoherence correction is turned off by setting `decoh_alpha=0`. However, this value corresponds rather to an infinitely strong decoherence correction. Turning it off means setting it to infinity. This is quite a confusing feature. We should create a new keyword for turning on/off the decoherence correction. When someone uses `decoh_alpha=0`, ABIN should issue an error since this would lead to a divergence.

JanosJiri:


JanosJiri:
If you approve @danielhollas, I will modify the manual accordingly.

PR #251, opened 2026-06-22
Decoupling of temp0 and temp keywords

### LINKED ISSUE #246: Decouple `temp` and `temp0` keywords
Currently, if `temp0` is not set and velocities are not provided, `temp` controls the generation of initial velocities. However, this often causes problems because the simulation will run even if you accidentally forget to provide the velocities and takes the temperature from `temp`, which many people left assigned in the thermostat section. We should decouple the two keyword and if `temp0` is not set and velocities are not provided, we should issue an error. `temp` should not be used for generating initial velocities.

JanPolena:
Setting the `temp0` keyword within the `input.in` file is now mandatory, unless the initial velocities are provided either from an external file or by restarting the calculation. In other cases, the user gets an error. 

As such, I also had to **change the test files** so that they contain the `temp0` keyword.

In addition, I have created a new test in the `INIT` directory to test forgetting of the `temp0` keyword in `input.in` when initial velocities are not provided.


danielhollas:
Thanks for the PR and sorry for the delay in review. I will get to it next week, but on a quick skim this looks great!

PR #252, opened 2026-06-30
Temporarily skip unit tests on macos runner

danielhollas:
For some reason the unit tests on macos-26 runners started failing with a linker error. Let's skip them for now.

```
ld: file cannot be open()ed, errno=2 (No such file or directory)
path=/opt/homebrew/Cellar/gcc/15.2.0_1/lib/gcc/current/libgomp.dylib in
'/opt/homebrew/Cellar/gcc/15.2.0_1/lib/gcc/current/libgomp.dylib'
```


JanosJiri (approved):
Yes, sounds reasonable to me.

PR #254, opened 2026-07-06
Remove shellcheck pre-commit hook

danielhollas:
Unfortunately, shellcheck (BASH linter) pre-commit hook requires Docker for installation, which many people will not have. Let's remove it for now. One can still [install shellcheck](https://github.com/koalaman/shellcheck#user-content-installing) using e.g. apt and run it manually.

### DIFF
--- a/prek.toml
+++ b/prek.toml
@@ -25,17 +25,6 @@ hooks = [
   }
 ]
 
-[[repos]]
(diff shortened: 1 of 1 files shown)


JanosJiri (approved):
That's a pitty..

PR #255, opened 2026-07-06
egrep is deprecated, use grep -E instead

danielhollas:
On my Fedora 39 system (already a few years old), using egrep prints a deprecation warning

```
❯ bash -c "egrep cds README.md" 
egrep: warning: egrep is obsolescent; using grep -E
```


JanosJiri (approved):
👍

PR #258, opened 2026-08-14
feat: MACE logging (reopened)

FelyCZ:
This PR replaces the standard use of 'print' with a correctly configured 'logging' module.

It also introduces a dump of the environment information (for debugging purposes).

### DIFF
--- a/interfaces/MACE/mace_server.py
+++ b/interfaces/MACE/mace_server.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 # /// script
-# requires-python = ">=3.8"
+# requires-python = ">=3.9"
 # dependencies = [
 #     "ase>=3.18.0",
 #     "mace-torch>=0.3.10",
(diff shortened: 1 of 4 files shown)


FelyCZ:
Reopened & reverted commit with the code highlighting change.

interfaces/MACE/mace_server.py
```diff
@@ -18,16 +19,79 @@
...
+    level = logging.DEBUG if debug else logging.INFO
+    logging.basicConfig(
+        level=level,
+        format="%(asctime)s | %(levelname)-8s | %(message)s",
+        datefmt="%Y-%m-%d %H:%M:%S",
+        stream=sys.stdout,
+        force=True,
+    )
+    logging.captureWarnings(True)
+    warnings.formatwarning = lambda msg, cat, fname, lineno, line=None: (
```
danielhollas, line 45: Could you please explain these two lines a bit (mainly a motivation behind them)

interfaces/MACE/mace_server.py
```diff
@@ -18,16 +19,79 @@
...
+        f"{fname}:{lineno}: {cat.__name__}: {str(msg).strip()}"
+    )
+
+
+# Configure logger at module import time (debug=True by default)
+setup_logger(debug=True)
+
+
+def log_environment_info(config):
```
danielhollas, line 54: Nice, this is great. I am quite tempted to just print this by default, because it is very useful and we only print it once at the top.

interfaces/MACE/mace_server.py
```diff
@@ -160,24 +224,26 @@ def connect_to_abin():
...
+    logger.info("Connection from ABIN accepted!")
     return port_name, abin_comm
 
 
 # https://docs.python.org/3/library/sys.html#sys.excepthook
 def exception_handler(shutdown_callback, exception_type, exception, traceback):
     """Try to gracefully shutdown communication with ABIN upon uncaught exceptions"""
-    print(f"Unexpected {exception_type.__name__}: {exception}")
-    print_tb(traceback)
+    logger.error(
```
danielhollas, line 243: I think we should be using `logger.exception` here, this is exactly its purpose, and then we can drop the manual printing of stack traces. Can you try and test it?

https://docs.python.org/3/library/logging.html#logging.Logger.exception

interfaces/MACE/mace_server.py
danielhollas: Also, if we change this to info level, we can drop the newly added `--debug`, at least for now.

danielhollas (changes requested):
Thank you, this looks great! Just a couple of small suggestions.

interfaces/MACE/mace_server.py
FelyCZ: > Could you please explain these two lines a bit (mainly a motivation behind them)

Sure thing! This is basically a warning interceptor. When any upstream module uses Python's `warning.warn()` method, this part of MACE server captures it and handles it. What it allows here is to have all upstream warnings formatted the same way as our log is.
TL;DR first line enables capture of warnings, second tells Python how to format it before logging using our logger.

### Example
This is with `captureWarnings` enabled:
```
2026-08-06 01:25:10 | INFO     | Received atom types: ['O', 'H', 'H']
2026-08-06 01:25:10 | WARNING  | /home/jakub/abin/.venv/lib/python3.13/site-packages/e3nn/o3/_wigner.py:10: UserWarning: Environment variable TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD detected, since the`weights_only` argument was not explicitly passed to `torch.load`, forcing weights_only=False.
cuequivariance or cuequivariance_torch is not available. Cuequivariance acceleration will be disabled.
2026-08-06 01:25:11 | DEBUG    | matplotlib data path: /home/jakub/abin/.venv/lib/python3.13/site-packages/matplotlib/mpl-data
2026-08-06 01:25:11 | DEBUG    | CONFIGDIR=/home/jakub/.config/matplotlib
```

And this how it would look without it.
```
2026-08-06 01:25:10 | INFO     | Received atom types: ['O', 'H', 'H']
/home/jakub/abin/.venv/lib/python3.13/site-packages/e3nn/o3/_wigner.py:10: UserWarning: Environment variable TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD detected, since the`weights_only` argument was not explicitly passed to `torch.load`, forcing weights_only=False.
cuequivariance or cuequivariance_torch is not available. Cuequivariance acceleration will be disabled.

2026-08-06 01:25:11 | DEBUG    | matplotlib data path: /home/jakub/abin/.venv/lib/python3.13/site-packages/matplotlib/mpl-data
2026-08-06 01:25:11 | DEBUG    | CONFIGDIR=/home/jakub/.config/matplotlib
```

interfaces/MACE/mace_server.py
FelyCZ: It is printed by default - as `--debug=True` by default. Providing `--debug` or `--debug false` flag when running the server toggles the state to `False`.
I've made it debug level, so it can be easily disabled.

interfaces/MACE/mace_server.py
FelyCZ: tested - we cannot use it directly.

`logger.exception()` is intended to be used inside `except` block, because calling it from this `exception_handler` method would result in `exc_info=(None, None, None)` as it is outside the exception call stack.

two options are available:
1. keep it this way
2. wrap the server loop inside `try...except` block and call `logger.exception` from there.

I propose **option 1.** as it results in minimal diff and preserves the full traceback.

